PR #668, opened 2022-02-25
MGDSTRM-7464 switching the ingress sizing to be based on demand

shawkins:
Replacement for #666 

also increasing the memory request - based upon the observation of memory usage in production.

This switches the ingress logic from being node based to being demand based. There are pluses and minuses for doing it this way. The best plus is that it's more based upon actual need - that becomes more relevant as we need a greater number of ingress replicas per managed kafka (when the ingress/egress goes up). The other benefit is that you don't need to hard code what is the potential maximum load per node, instead you characterize what each ingress replica can handle and determine how many you need.

There are also 2 new properties that simply allow you to override how many replicas you actually want - that should be useful to kas-install and instance profiling.

The minuses are:

- it's more complicated, but yet still has to make simplifying assumptions such as assuming that the demand will be equal in all zones
- it may not be worst case enough. Accounting for the case where an ingress replica ends up on the same node as a broker is currently modeled as half broker traffic is coming from the ingress replica that is next to it (approximately the case when there are 2 replicas per zone). As you add replicas though a smaller proportion of that brokers traffic is likely going though it's collocated replica.
- ~~it will more rapidly scale down the number of replicas than the node based strategy. Scaling down a replica may force clients to reconnect (does anyone know for sure what the behavior is?)~~ the logic will conservatively scale down only when you have 3 excess replicas or when nodes are scaled down.
- it may need a tighter reconciliation than 3 minutes (if we ever support changing the ingress/egress capacity on the fly)
- it does not guard against instances being created without quotas or with unrealistic values - those scenarios (instance profiling) would use the overrides
Any feedback is welcome.

Other thoughts:

- we can support a far greater number of connections per ingress replica, but we have not validated what the memory requirement will be.
- we'll eventually likely want to set the resource requirements differently for the broker and default replicas.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/InformerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/InformerManager.java
@@ -29,6 +29,7 @@ import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Stream;
 
@@ -170,6 +171,13 @@ public class InformerManager {
                 this.eventSource.onUpdate(k, k);
             });
         }
+    }
+
+    public List<Kafka> getKafkas() {
+        if (kafkaInformer != null) {
+            return kafkaInformer.getList();
+        }
+        return Collections.emptyList();
     }
 
     public void resyncManagedKafka() {
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -401,7 +401,7 @@ public class IngressControllerManager {
             IngressController existing, int replicas, LabelSelector routeSelector, String topologyValue) {
 
         Optional<IngressController> optionalExisting = Optional.ofNullable(existing);
-        IngressControllerBuilder builder = optionalExisting.map(IngressControllerBuilder::new).orElse(new IngressControllerBuilder());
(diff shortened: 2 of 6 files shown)


shawkins:
@grdryn can you review again with the changes to limit how often we'll scale the ingress replicas down.

k-wall (approved):

MikeEdgar (approved):
Just a trivial suggestion

shawkins:
@rareddy @grdryn @k-wall One more iteration.  Based upon all concerns going around about node costs it seems best if this code change would not introduce any additional node requirements.  So I'm adding a factor to state what percentage of the peak throughput you are actually trying to satisfy.  I settled on 40% vs the initially proposed 50% because that matches the ingress/egress scaling factor from GA to the currently proposed SU.  This roughly works out 2 ingress replicas for up to 34 3 broker instances 4 replicas for up to 52 3 broker instances, etc.  If this is applied with GA deployments, then we'll simply use 2 replicas up to 52+ (it will be slightly higher) 3 broker instances.  I am okay with this potential reduction given that we're also ensuring greater resource availability.

rareddy:
@shawkins node count wise what you see for 2 replicas vs 4?

shawkins:
> @shawkins node count wise what you see for 2 replicas vs 4?

I'll be updating the proposed node formula on the issue as well.  2 replicas will be accounted for already in whatever is used as the base amount of additional nodes.  Beyond that we're going to assume that one set of replicas needs an additional node, so 3, 4, 5 would all need 3 additional nodes.  Beyond that we currently aren't considering.  So basically we're right back to using either +12 or +15 depending upon the amount of brokers.

PR #669, opened 2022-03-01
Expose instance metrics

k-wall:
Exposes some 'limit' type metrics on a per kafka instance basis.  Will help allow the system overall to adapt to support more than one "profile".

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
@@ -30,7 +30,13 @@ public class MetricsManager {
     static final String KAFKA_INSTANCE_MAX_MESSAGE_SIZE_LIMIT = "kafka_instance_max_message_size_limit";
     static final String KAFKA_INSTANCE_CONNECTION_LIMIT = "kafka_instance_connection_limit";
     static final String KAFKA_INSTANCE_CONNECTION_CREATION_RATE_LIMIT = "kafka_instance_connection_creation_rate_limit";
-    static final Tag OWNER = Tag.of("owner", "KafkaInstanceMetricsManager");
+
+    static final String TAG_LABEL_OWNER = "owner";
+    static final String TAG_LABEL_BROKER_ID = "broker_id";
+    static final String TAG_LABEL_NAMESPACE = "namespace";
+    static final String TAG_LABEL_INSTANCE_NAME = "instance_name";
+
+    static final Tag OWNER = Tag.of(TAG_LABEL_OWNER, "KafkaInstanceMetricsManager");
 
     @Inject
     ResourceInformerFactory resourceInformerFactory;
@@ -72,7 +78,7 @@ public class MetricsManager {
 
                 if (kafka.getSpec() != null && kafka.getSpec().getKafka() != null) {
                     IntStream.range(0, kafka.getSpec().getKafka().getReplicas()).forEach(ordinal -> {
-                        Tags brokerTags = Tags.concat(tags, Tags.of(Tag.of("broker", String.valueOf(ordinal))));
+                        Tags brokerTags = Tags.concat(tags, Tags.of(Tag.of(TAG_LABEL_BROKER_ID, String.valueOf(ordinal))));
 
                         if (kafka.getSpec().getKafka().getListeners() != null) {
                             kafka.getSpec().getKafka().getListeners().stream().filter(l -> "external".equals(l.getName())).forEach(l -> {
@@ -87,7 +93,7 @@ public class MetricsManager {
 
             private Tags buildKafkaInstanceTags(Kafka obj) {
                 ObjectMeta metadata = obj.getMetadata();
-                return Tags.of(Tag.of("namespace", metadata.getNamespace()), Tag.of("name", metadata.getName()), OWNER);
+                return Tags.of(Tag.of(TAG_LABEL_NAMESPACE, metadata.getNamespace()), Tag.of(TAG_LABEL_INSTANCE_NAME, metadata.getName()), OWNER);
             }
 
             private Double replicas(Kafka k) {
--- a/operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
@@ -22,7 +22,7 @@ import static org.bf2.operator.operands.AbstractKafkaCluster.EXTERNAL_LISTENER_N
 
 @Startup
 @ApplicationScoped
-public class MetricsManager {
+public class MetricsManager implements ResourceEventHandler<Kafka>{
     static final String KAFKA_INSTANCE_SPEC_BROKERS_DESIRED_COUNT = "kafka_instance_spec_brokers_desired_count";
     static final String KAFKA_INSTANCE_PARTITION_LIMIT = "kafka_instance_partition_limit";
     static final String KAFKA_INSTANCE_MAX_MESSAGE_SIZE_LIMIT = "kafka_instance_max_message_size_limit";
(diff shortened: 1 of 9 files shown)


k-wall:
The SBM test failure is caused by the refactoring of the ResourceInformers produced by MockResourceInformerFactory being enhanced to actually emit events.  This broke some assumptions made by that test.  That is cleaned up by #673.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -0,0 +1,111 @@
...
+                                meterRegistry.gauge(KAFKA_INSTANCE_CONNECTION_CREATION_RATE_LIMIT, listenerTags, l, this::maxConnectionCreationRate);
+                            });
+                        }
+                    });
+                }
+            }
+
+            private Tags buildKafkaInstanceTags(Kafka obj) {
+                ObjectMeta metadata = obj.getMetadata();
+                return Tags.of(Tag.of("namespace", metadata.getNamespace()), Tag.of("name", metadata.getName()), OWNER);
```
grdryn, line 90: I guess tags are what Prometheus calls labels? I see a lot of metrics have a `strimzi_io_name` label rather than a `name`. I wonder if there's any reason for that namespacing of label names that we'd need to be aware of?

grdryn (approved):

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -0,0 +1,111 @@
...
+                meterRegistry.gauge(KAFKA_INSTANCE_BROKERS_DESIRED_COUNT, tags, kafka, this::replicas);
+                meterRegistry.gauge(KAFKA_INSTANCE_PARTITION_LIMIT, tags, kafka, k -> kafkaConfigValue(k, KafkaCluster.MAX_PARTITIONS));
+                meterRegistry.gauge(KAFKA_INSTANCE_MAX_MESSAGE_SIZE_LIMIT, tags, kafka, k -> kafkaConfigValue(k, KafkaCluster.MESSAGE_MAX_BYTES));
+
+                if (kafka.getSpec() != null && kafka.getSpec().getKafka() != null) {
+                    IntStream.range(0, kafka.getSpec().getKafka().getReplicas()).forEach(ordinal -> {
+                        Tags brokerTags = Tags.concat(tags, Tags.of(Tag.of("broker", String.valueOf(ordinal))));
+
+                        if (kafka.getSpec().getKafka().getListeners() != null) {
+                            kafka.getSpec().getKafka().getListeners().stream().filter(l -> "external".equals(l.getName())).forEach(l -> {
```
MikeEdgar, line 78: Consider a constant for `"external"` since it's shared between this and `AbstractKafkaCluster`.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
k-wall: Metric labels that begin `strimzi_io` originate from the pod's labels of pods controlled by strimzi.  

```
oc get pods penguin-kafka-0 --show-labels
NAME              READY   STATUS    RESTARTS   AGE     LABELS
penguin-kafka-0   1/1     Running   0          5h14m   app.kubernetes.io/instance=penguin,app.kubernetes.io/managed-by=strimzi-cluster-operator,app.kubernetes.io/name=kafka,app.kubernetes.io/part-of=strimzi-penguin,controller-revision-hash=penguin-kafka-5d6fd54bcc,ingressType=sharded,managedkafka.bf2.org/kas-multi-zone=true,managedkafka.bf2.org/kas-us-east-1a=true,managedkafka.bf2.org/kas-us-east-1b=true,managedkafka.bf2.org/kas-us-east-1c=true,statefulset.kubernetes.io/pod-name=penguin-kafka-0,strimzi.io/cluster=penguin,strimzi.io/kind=Kafka,strimzi.io/name=penguin-kafka
```

these get mapped into labels by this  [label rewrite rule](https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/blob/6f333220ab8ba5eeccf72373f63900cacad6108c/resources/prometheus/pod_monitors/kafka-metrics.yaml#L20) in the pod monitor into what you see in prometheus.

Should labels be namespaced?  I would argue no, doing so would make joining prometheus metrics produced by different components really hard.  You end up with PromQL full of [`label_replaces()`](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace) or record rules that map labels to a common domain.  See what I am doing https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/pull/165 to get to a common set of label names.  Comments/thoughts welcome!

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -540,6 +543,7 @@ private ResourceRequirements buildKafkaExporterResources(ManagedKafka managedKaf
         config.put("inter.broker.protocol.version", this.kafkaManager.currentKafkaIbpVersion(managedKafka));
         config.put("ssl.enabled.protocols", "TLSv1.3,TLSv1.2");
         config.put("ssl.protocol", "TLS");
+        config.put(MESSAGE_MAX_BYTES, 1048588); // Kafka's default
```
shawkins, line 546: Please combine with https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/8944fcc3aec896b750e70b28b2727881eaad4d73/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java#L561

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Locale;
+import java.util.stream.IntStream;
+
+import static org.bf2.operator.managers.InformerManager.filter;
+import static org.bf2.operator.operands.AbstractKafkaCluster.EXTERNAL_LISTENER_NAME;
+
+@Startup
+@ApplicationScoped
+public class MetricsManager {
+    static final String KAFKA_INSTANCE_BROKERS_DESIRED_COUNT = "kafka_instance_brokers_desired_count";
```
shawkins, line 30: We can have two different desired broker counts, this one and the one that we infer from the capabilities.  We may me need more precise naming here or in the other pr about what desired means.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -0,0 +1,121 @@
...
+    @Inject
+    KubernetesClient kubernetesClient;
+
+    private volatile ResourceInformer<Kafka> kafkaResourceInformer;
+
+    protected synchronized void createInformer() {
+        if (kafkaResourceInformer != null) {
+            return;
+        }
+        kafkaResourceInformer = resourceInformerFactory.create(Kafka.class, filter(kubernetesClient.resources(Kafka.class, KafkaList.class)), new ResourceEventHandler<>() {
```
shawkins, line 59: With some refactoring we should be able to this handler to the existing informer, so that we don't have to maintain a separate cache/watch.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: Done

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
k-wall: For clarity let me say, for this use-case, I need to publish the number of brokers that the kafka instance is supposed to have according to its configuration.  If an broker or an AZ happens to fail so that number of brokers is fewer, the user of the metric can tell this by comparing this desired value with the actual value reported by the kafka exporter.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
k-wall: I spent a while looking at whether i could layer responsibilities into `ManagedKafkaController`, but I didn't really see the way forward.  I'm not too familiar with code, so I could well be missed better solutions.  I do like the separation of concerns that the current solution brings.

I'm open to refactoring what I have but I'll need some guidance.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -0,0 +1,121 @@
...
+                meterRegistry.gauge(KAFKA_INSTANCE_MAX_MESSAGE_SIZE_LIMIT, tags, kafka, k -> kafkaConfigValue(k, KafkaCluster.MESSAGE_MAX_BYTES));
+
+                if (kafka.getSpec() != null && kafka.getSpec().getKafka() != null) {
+                    IntStream.range(0, kafka.getSpec().getKafka().getReplicas()).forEach(ordinal -> {
+                        Tags brokerTags = Tags.concat(tags, Tags.of(Tag.of(TAG_LABEL_BROKER_ID, String.valueOf(ordinal))));
+
+                        if (kafka.getSpec().getKafka().getListeners() != null) {
+                            kafka.getSpec().getKafka().getListeners().stream().filter(l -> EXTERNAL_LISTENER_NAME.equals(l.getName())).forEach(l -> {
+                                String listenerName = String.format("%s-%d", l.getName().toUpperCase(Locale.ROOT), l.getPort());
+                                Tags listenerTags = Tags.concat(brokerTags, Tags.of(Tag.of(TAG_LABEL_LISTENER, listenerName)));
```
MikeEdgar, line 89: Maybe a small comment here to give the reason for the `toUpperCase` since it's outside of fleetshard.

MikeEdgar (approved):

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
shawkins: It's more about the terminology - we have two different desired replica counts.  Until we do quite a lot more work if this situation arises they will not be eventually consistent in any immediate sense.  I vote either for calling this one spec brokers, or changing the other other annotation to inferred or required brokers.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
k-wall: Done

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
k-wall: @shawkins I struggled a fair bit  with this refactoring.  The difficulty was that the kafka informer is created late (only once the operate knows the CRDs are present).  That meant I needed some mechanism to trigger the registration of the handler with the kafka informer, which lead me to `org.bf2.common.ResourceInformerFactory#addEventHandler`. I'm okay with the production code.

I'm not so happy with MockResourceInformerFactory - I've made the code worse.  I think MockResourceInformerFactory would be better eliminated, but some of the existing tests seem to rely on subtleties of the mock's behaviour (IngressControllerManagerTest).   Refactoring the tests that quickly got beyond what I was happy with for this PR, so I would prefer remove the dependency on a separate PR. 

I did try some alternative paths: using `QuarkusMock.installMockForType` for ResourceInformerFactory, but it seems to get confused by the presence of the MockResourceInformerFactory. 

WDYT?

shawkins: > I'm not so happy with MockResourceInformerFactory - I've made the code worse.

My original intention was to disable all async event processing - whether scheduled or something that would be handled by an informer, so that tests could be written based upon just single threaded behavior.  

The addition of:

if (eventHandler != null) {
   watchListDeletable.inform((ResourceEventHandler) eventHandler);
}

Definitely violates that as that is starting up a real informer - I misread the pr initially and should have said something earlier.

I don't mind refactoring to remove MockResourceInformerFactory entirely.  The intermediate step is to make sure the mock server state is reset as necessary.

But the goal of no-async behavior is very difficult when starting a full quarkus - beans may be being started in parallel and if they are all making updates to the same mock store, then you may see things you don't expect.  When we upgrade quarkus we should be able to remove the startup annotation from a lot of bean - lazy initialization will make things a bit better.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -598,6 +601,9 @@ private ResourceRequirements buildKafkaExporterResources(ManagedKafka managedKaf
         // custom authorizer configuration
         addKafkaAuthorizerConfig(managedKafka, config);
 
+        if (managedKafka.getSpec().getCapacity().getMaxPartitions() != null) {
+            config.put(MAX_PARTITIONS, managedKafka.getSpec().getCapacity().getMaxPartitions());
+        }
```
MikeEdgar, line 606: Requires update to test file `/expected/custom-config-strimzi.yml`

shawkins (approved):
A minor simplification would be to have MetricsManager directly implement ResourceEventHandler, so that you can call the methods directly in your tests without scraping the handler from the mock InformerManager.  Your can of course still verify that registerKafkaInformerHandler was called.

k-wall:
> A minor simplification would be to have MetricsManager directly implement ResourceEventHandler, so that you can call the methods directly in your tests without scraping the handler from the mock InformerManager.

Nice.  That tidies things up a lot.

> Your can of course still verify that registerKafkaInformerHandler was called.

I tried that and fell at the first try.  This works providing test registersWithInformerManager runs first!  I guess the mock is being reset between test invocations.

```
    @InjectMock
    InformerManager informerManager;

    @Test
    public void registersWithInformerManager() {
        metricsManager.toString();
        verify(informerManager, times(1)).registerKafkaInformerHandler(isA(MetricsManager.class));
    }
```

I also tried combinations using `QuarkusMock.installMockForType`.  I also tried putting the verification in a postConstruct, that failed too.

I gave up.

MikeEdgar (approved):
lgtm

shawkins:
> I guess the mock is being reset between test invocations.

Sigh, as we'd like for all bean...

> I'm happy to leave it out.

Yes, it's not imperative to test all of the wiring logic.

PR #673, opened 2022-03-03
Refactor StrimziBundleManager unit tests to avoid the need for 'test-only' methods.

k-wall:
The SBMTest failures were provoked by a changes to the MockResourceInformerFactory.  The factory was changed to produce informers that actually emit events - which I found useful elsewhere. This broke some assumptions made by the SBMTests.

The tests previously relied on testing business logic by calling methods exposed specifically for the unit test.  The code paths where the watcher actually called the business logic were untested making for a somewhat weak test.

The test is now refactored to use a simple watcher that mimics the important actions that would be made by OLM.  This
allows the code to be tested under conditions that match life much more closely.

### DIFF
--- a/operator/pom.xml
+++ b/operator/pom.xml
@@ -85,6 +85,11 @@
             <artifactId>quarkus-test-kubernetes-client</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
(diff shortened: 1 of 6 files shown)


shawkins (approved):
LGTM

MikeEdgar (approved):

PR #675, opened 2022-03-04
using partitions to drive the selection of the number of brokers

shawkins:
This switches the fleetshard operator to using the partition count to drive the broker selection.

This requires adding a physical partition capacity value to the configuration to use as a reference.  I removed the old replicas value from the config, but it may be a good idea to add that back as an override to simply testing as noted in the ManagedKafkaProvisioner.

To support a single broker a subsequent pr will externalize the scalingAndReplicaionFactor - at worst it could be separate properties.

My understanding is that changing the broker count after the fact is not something that we don't want to do yet.  So similar to the consideration of not reducing volume sizes, to get the replica count we have to use both the managedkafka and the kafka, which makes these changes seem a bit messy.

This pr is not yet accounting for the physical limits of the other dimensions - we are assuming that the initial values will be supported based upon whatever the partition count is.  If/when we add that we'll also need to update the ManagedKafka status / status capacity when there is a discrepancy.

This does highlight the inter-relatedness of all the values - you cannot for example later increase the physical support for partitions and hope to see a reduction in the number of physical units - the physical support for each dimension (ingress/egress, connections) must similarly increase.  Related to this if we change the number of brokers the total storage must either stay the same or increase - so if we are satisfying something with 1 physical unit, and need that to be 2 physical units for whatever reason, we'll have double the storage of the ManagedKafka capacity value.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -89,7 +89,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -115,6 +114,11 @@ public class KafkaCluster extends AbstractKafkaCluster {
     private static final String KAFKA_EXPORTER_LOG_LEVEL = "logLevel";
 
     private static final String DIGEST = "org.bf2.operator/digest";
+    /* tracks the number of brokers desired as inferred from the capacity
+     * it may differ from the actual number of brokers on the kafka resource
+     * and will later need a reconciliation process
+     */
+    private static final String REPLICAS = "org.bf2.operator/desired-broker-replicas";
     private static final String IO_STRIMZI_KAFKA_QUOTA_STATIC_QUOTA_CALLBACK = "io.strimzi.kafka.quotas.StaticQuotaCallback";
 
     private static final String SERVICE_ACCOUNT_KEY = "managedkafka.kafka.acl.service-accounts.%s";
@@ -197,22 +201,26 @@ public class KafkaCluster extends AbstractKafkaCluster {
     public Kafka kafkaFrom(ManagedKafka managedKafka, Kafka current) {
         KafkaBuilder builder = current != null ? new KafkaBuilder(current) : new KafkaBuilder();
 
+        int actualReplicas = getBrokerReplicas(managedKafka, current);
+        int desiredReplicas = getBrokerReplicas(managedKafka, null);
+
         KafkaBuilder kafkaBuilder = builder
                 .editOrNewMetadata()
                     .withName(kafkaClusterName(managedKafka))
                     .withNamespace(kafkaClusterNamespace(managedKafka))
                     .withLabels(buildKafkaLabels(managedKafka))
                     .withAnnotations(buildKafkaAnnotations(managedKafka, current))
+                    .addToAnnotations(REPLICAS, String.valueOf(desiredReplicas))
                 .endMetadata()
                 .editOrNewSpec()
                     .editOrNewKafka()
                         .withVersion(this.kafkaManager.currentKafkaVersion(managedKafka))
                         .withConfig(buildKafkaConfig(managedKafka, current))
-                        .withReplicas(this.config.getKafka().getReplicas())
+                        .withReplicas(actualReplicas)
                         .withResources(buildKafkaResources(managedKafka))
                         .withJvmOptions(buildKafkaJvmOptions(managedKafka))
                         .withStorage(buildKafkaStorage(managedKafka, current))
-                        .withListeners(buildListeners(managedKafka))
+                        .withListeners(buildListeners(managedKafka, actualReplicas))
                         .withRack(buildKafkaRack(managedKafka))
                         .withTemplate(buildKafkaTemplate(managedKafka))
                         .withMetricsConfig(buildKafkaMetricsConfig(managedKafka))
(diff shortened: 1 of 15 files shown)


rareddy:
Looks good.

> My understanding is that changing the broker count after the fact is not something that we don't want to do yet.

I believe we need to wait until cruise control's availability for this.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -645,36 +673,36 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current) {
     /**
      * Get per broker value
      */
-    private long getPerBrokerBytes(ManagedKafka managedKafka, Function<ManagedKafka, Quantity> capacityValue, Supplier<String> defaultValue) {
+    private long getPerBrokerBytes(ManagedKafka managedKafka, Kafka current, Function<ManagedKafka, Quantity> capacityValue, Supplier<String> defaultValue) {
```
tombentley, line 676: Why do we pass a Function here if we just call it immediately? Why not just pass the value directly? Same with the `Supplier`.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -111,6 +111,7 @@
     private static final String KAFKA_EXPORTER_LOG_LEVEL = "logLevel";
 
     private static final String DIGEST = "org.bf2.operator/digest";
+    private static final String REPLICAS = "org.bf2.operator/replicas";
```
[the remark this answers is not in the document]
shawkins: yes, I'll update that.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -238,6 +243,29 @@ public Kafka kafkaFrom(ManagedKafka managedKafka, Kafka current) {
...
     }
 
+    @Override
+    public int getReplicas(ManagedKafka managedKafka) {
+        return getBrokerReplicas(managedKafka, cachedKafka(managedKafka));
+    }
+
+    public int getBrokerReplicas(ManagedKafka managedKafka, Kafka current) {
+        Integer maxPartitions = managedKafka.getSpec().getCapacity().getMaxPartitions();
+        int scalingAndReplicaionFactor = 3;
```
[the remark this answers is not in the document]
shawkins: sure

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
shawkins: We certainly can just pass the value.  In a different version of the method the default was optional, it can be refactored again to work that way.

shawkins:
> Can you elaborate on this, I'm not sure I follow.

If you are satisfying the user request with 1 physical unit (assuming that is 3 brokers), it will start of with roughly maxRetention / 3 storage at each broker.  If you have to switch to 2 physical units for whatever reason to satisfy the same user request, the volumes will still have maxRetention / 3 because shrinking is not supported but there will be 6 of them.

tombentley:
Thanks! Where does the shrinking not supported restriction come from?

shawkins:
> Thanks! Where does the shrinking not supported restriction come from?

Strimzi - https://strimzi.io/docs/operators/in-development/configuring.html "Decreasing the size of existing persistent volumes is not possible."

tombentley:
Hmm. Turns out Strimzi inherits it from Kube, and it's not supported in EBS either, apparently. This will be a problem when we support instance scale up, which I hadn't really realised before now. :disappointed:

shawkins:
> Hmm. Turns out Strimzi inherits it from Kube, and it's not supported in EBS either, apparently. 

Yes, it's not a built-in operation.  Generally you have to create a smaller volume, stop writes to the volume to shrink, fully sync to the smaller volume, then start using the smaller volume.

> This will be a problem when we support instance scale up, which I hadn't really realised before now. disappointed

With 1 SU = 1TB and 2 SU = 2TB and so forth there isn't an issue as we're going to keep the same amount of storage per broker.  But with custom storage amounts there is a problem - 1 SU with 5 TB cannot simply upgrade to 2 SU with 5 TB.

rareddy (approved):
lgtm

operator/src/main/resources/application.properties
```diff
@@ -126,15 +126,28 @@ topic=*;operations=describe;level=DEBUG \n\
...
 managedkafka.kafka.jvm-xms=6442450944
 
+# absolute capacity specific to the above
+
+# this is meant as a pre-replication per broker value
+#
+# the spec capacity value is understood to be
+# X fully replicated partitions typically with replication factor 3
+managedkafka.kafka.partition-capacity=1500
+
```
MikeEdgar, line 150: I'm not following what this property defines (versus the MK's `maxPartitions`). Is this saying that there can be 500 partitions with replication factor 3?

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -242,6 +250,29 @@ public Kafka kafkaFrom(ManagedKafka managedKafka, Kafka current) {
...
+
+    public int getBrokerReplicas(ManagedKafka managedKafka, Kafka current) {
+        Integer maxPartitions = managedKafka.getSpec().getCapacity().getMaxPartitions();
+        int scalingAndReplicationFactor = 3;
+        int desiredReplicas = scalingAndReplicationFactor;
+        if (maxPartitions != null) {
+            double physicalCapacity = config.getKafka().getPartitionCapacity();
+            desiredReplicas = (int) Math.ceil(maxPartitions * scalingAndReplicationFactor / physicalCapacity);
+            // round to only even multiples
+            desiredReplicas = (int) (Math.ceil(desiredReplicas / (double)scalingAndReplicationFactor) * scalingAndReplicationFactor);
```
MikeEdgar, line 266: This comment isn't matching up with my parsing of the calculation. For the current configuration (assuming `1000` max partitions from control plane), I think this updates `desiredReplicas` from 2 to 3. What am I missing?

operator/src/main/resources/application.properties
shawkins: It is supposed to be a physical limit supported per broker supported by this configuration / profile.

The MK maxPartitions is a user request for total partitions = maxPartitions * replication factor.

We use the physical limit to determine how many brokers are needed relative to that user request.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
shawkins: > I think this updates desiredReplicas from 2 to 3. What am I missing?

I don't know that you are missing anything.  It will round from 2 to 3 as we only want even multiples of the scalingAndReplicationFactor.  If we determine that 4 brokers are needed, we'll then round to 6 and so forth.

MikeEdgar: Ah, I was missing that we want even multiples "of the scalingAndReplicationFactor" rather than simply "even".

MikeEdgar (approved):

PR #678, opened 2022-03-07
Set props needed for authorizer partition limiting

grdryn:


k-wall (approved):
lgtm

k-wall:
This is backwardly compatible?  I mean an older version of the plugin will ignore the new settings.  Right?

shawkins (approved):
LGTM - just need to update the expected files.

grdryn:
@k-wall I've added some additional properties here, based on your suggestions in https://github.com/bf2fc6cc711aee1a0c2a/kafka-group-authorizer/pull/21. I've also prefixed the `max.partitions` one to be similar to the others.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
```diff
@@ -390,6 +392,14 @@ public LoggingSuppressionWindow getLoggingSuppressionWindow() {
         public void setLoggingSuppressionWindow(LoggingSuppressionWindow loggingSuppressionWindow) {
             this.loggingSuppressionWindow = loggingSuppressionWindow;
         }
+
+        public String getPrivatePrefix() {
```
SamBarker, line 396:
```suggestion
        public String getPrivateTopicPrefix() {
```
or
```suggestion
        public String getInternalTopicPrefix() {
```
I was initially confused at to what `privatePrefix` meant then I spotted the property.  Not sure if `private` is the right term so maybe `internal` is better, I think that's what we have referred to it as in other contexts.

SamBarker (approved):
LGTM

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
MikeEdgar: I'm not opposed to a name change, but note that this prefix applies to other "internal" objects besides topics like transaction_id, consumer groups, and client IDs.

SamBarker: Ohh, I'd miss read the flow (and scope) of the change I thought this was being driven from `strimzi.authorization.custom-authorizer.partition-counter.private-prefix` but no its providing a value to it.

I was only intending to suggest the `partition-counter` called it a `*TopicPrefix`. 

The more I think about it less I like `internal` as a change from `private`. I thought it was being consistent with other usages such as the canary but I can't find any evidence of that on GitHub. So quite happy to stick with `private`.

operator/src/test/resources/expected/custom-config-strimzi.yml
```diff
@@ -121,6 +121,12 @@ spec:
       leader.imbalance.per.broker.percentage: 0
       connections.max.reauth.ms: 299000
       create.topic.policy.class.name: io.bf2.kafka.topic.ManagedKafkaCreateTopicPolicy
+      strimzi.authorization.custom-authorizer.partition-counter.max.partitions: 3000
+      strimzi.authorization.custom-authorizer.partition-counter.timeout-seconds: 10
+      strimzi.authorization.custom-authorizer.partition-counter.private-prefix: __redhat_
```
SamBarker, line 126:
```suggestion
      strimzi.authorization.custom-authorizer.partition-counter.private-topic-prefix: __redhat_
```
See https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/678/files#r822214537 for more discussion

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -598,6 +600,16 @@ private ResourceRequirements buildKafkaExporterResources(ManagedKafka managedKaf
         // custom authorizer configuration
         addKafkaAuthorizerConfig(managedKafka, config);
 
+        config.put("strimzi.authorization.custom-authorizer.partition-counter.max.partitions",
+                Optional.ofNullable(managedKafka.getSpec()).map(ManagedKafkaSpec::getCapacity).map(ManagedKafkaCapacity::getMaxPartitions).orElse(-1));
+        config.put("strimzi.authorization.custom-authorizer.partition-counter.timeout-seconds", 10);
+        config.put("strimzi.authorization.custom-authorizer.partition-counter.private-prefix",
```
rareddy, line 606: can we not create single property for this and `managedkafka.kafka.acl.private-prefix`?

grdryn: @rareddy I'm not 100% sure what you mean. This is picking up on the `managedkafka.kafka.acl.private-prefix` from the properties file here, and setting it on the config in the Kafka CR's `spec.kafka.config`. I don't see it otherwise being set anywhere there.

operator/src/test/resources/expected/custom-config-strimzi.yml
grdryn: Done, thanks!

grdryn:
I've updated this on top of Keith's changes (which added one of the properties that I was also going to set here). I've also added a new property for the schedule interval that the partition counter in the authorizer.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
rareddy: I mean I see the two properties defining the "prefix" can they not be combined into a single property?

grdryn: @rareddy I might still be misunderstanding, but I think there is only a single property defining the prefix (although it exists in two different contexts, with different names):

1. The fleetshard operator has a property called `managedkafka.kafka.acl.private-prefix` defined in the application.properties file. This isn't used directly in code (or passed on to other components as-is), it's referenced in other fleetshard properties in the application.properties, for substitution.

2. The group authorizer plugin will (as part of https://github.com/bf2fc6cc711aee1a0c2a/kafka-group-authorizer/pull/21) expect to know what the prefix is, but that isn't exposed to it through the broker configs anywhere (I suppose it could parse something like `strimzi.authorization.custom-authorizer.acl.010`, pull out and clean up the topic part, and use that, but that doesn't seem right). I instead chose to add a new property for it, and have named it with a similar prefix to the other properties where it's used: `strimzi.authorization.custom-authorizer.partition-counter.private-topic-prefix` as set in this PR.

So there's one fleetshard operator property, and one broker property (set to the value of the fleetshard operator property, but given a different name). Does that make sense, or am I still not getting what you're saying?

rareddy: ok, that is fine with me having different properties for these two contexts. Since they were both portraying the same issue I thought if a common property can be derived among both but does not seem that critical to have.

PR #680, opened 2022-03-08
externalizing scaling and replication factor for single broker case

shawkins:


rareddy:
Couple more changes I saw that are needed for Single AZ stuff are 
- ZK replica count (This goes along other configuration)
- Admins Server has default replica at 3, need to override by supplying an ENV var `KAFKA_ADMIN_REPLICATION_FACTOR` to whatever the Kafka replica size is.

shawkins:
> ZK replica count (This goes along other configuration)

There's already a comment about that in the config.

> Admins Server has default replica at 3, need to override by supplying an ENV var KAFKA_ADMIN_REPLICATION_FACTOR to whatever the Kafka replica size is.

Does it default to 3? https://github.com/bf2fc6cc711aee1a0c2a/kafka-admin-api/blob/d51b8c6c486a3ab0b93ccc890c43b560d9921af4/.env#L3

@MikeEdgar can you weigh in on the KAFKA_ADMIN_REPLICATION_FACTOR property

rareddy:
I was looking at https://github.com/rareddy/kafka-admin-api/blob/main/kafka-admin%2Fsrc%2Fmain%2Fjava%2Forg%2Fbf2%2Fadmin%2Fkafka%2Fadmin%2FTopicOperations.java#L31

which stopped me from creating a topic

MikeEdgar:
Yes, it has a default of 3 if not specified. Adding that to the admin deployment should resolve that.

shawkins:
Okay I'll update.  

There should probably be a change in the vert.x wrapping logic.  It is hiding that you can use a NewTopic constructor where the replication factor can be automatically inferred.

MikeEdgar (approved):
lgtm

k-wall:
Since this is merged this will be part of 0.20 when we cut the branch from main.  Added the milestone to reflect this.

PR #682, opened 2022-03-10
Correct the handling of metric updates.

k-wall:
I noticed that sometimes after creating a managedkafka, the new metrics would show a NaN value, rather than the reflection of the kafka resource as expected.  It turns out there is a race. I misunderstood how `meterRegistry.gauge()` works, if the metric already exists (same name and tags), the method is a no-op.  This meant metric continued to return values from the initial kaka object and any updates were ignored.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
@@ -134,8 +134,7 @@ public class MetricsManager implements ResourceEventHandler<Kafka>{
     private Optional<GenericKafkaListener> getExternalListener(AtomicReference<Kafka> r) {
         return getKafkaClusterSpec(r)
                 .map(KafkaClusterSpec::getListeners)
-                .orElse(List.of())
-                .stream().filter(l -> AbstractKafkaCluster.EXTERNAL_LISTENER_NAME.equals(l.getName())).findFirst();
+                .flatMap(ll -> ll.stream().filter(l -> AbstractKafkaCluster.EXTERNAL_LISTENER_NAME.equals(l.getName())).findFirst());
     }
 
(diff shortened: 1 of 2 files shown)


operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -43,68 +54,94 @@
...
+        externalListener.ifPresent(genericKafkaListener -> IntStream.range(0, kafka.getSpec().getKafka().getReplicas()).forEach(ordinal -> {
+            String listenerName = String.format("%s-%d", genericKafkaListener.getName().toUpperCase(Locale.ROOT), genericKafkaListener.getPort()); // Note - kafka itself capitalises the listener name.
+            Tag brokerTag = Tag.of(TAG_LABEL_BROKER_ID, String.valueOf(ordinal));
+            Tag listenerTag = Tag.of(TAG_LABEL_LISTENER, listenerName);
+            Tags listenerTags = tags.and(brokerTag, listenerTag);
+
+            orphanMeters.removeIf(m -> m.getId().getTags().contains(brokerTag));
+
+            meterRegistry.gauge(KAFKA_INSTANCE_CONNECTION_LIMIT, listenerTags, ref,
+                    r -> getExternalListener(r).map(GenericKafkaListener::getConfiguration).map(GenericKafkaListenerConfiguration::getMaxConnections).map(Integer::doubleValue).orElse(Double.NaN));
```
MikeEdgar, line 104: I'm new to this API - could you also pass `genericKafkaListener` (instead of `ref`) and skip the first step of `getExternalListener(r)` ?

MikeEdgar (approved):
lgtm. Just a question on the API usage.

grdryn (approved):

grdryn (approved):
LGTM

PR #686, opened 2022-03-14
Add scheme to `adminServerURI` in ManagedKafka status

MikeEdgar:
Include URI scheme (http/https) reported in the `adminServerURI` of the MK status.

MGDSTRM-7766


biswassri (approved):
This makes sense and looks good to me.

PR #687, opened 2022-03-16
[WIP] Factor out inline secrets from managedkafka CRs 

biswassri:

- bootstrap certificates
- oauth - clientid, clientSecret
- Canary service account


biswassri:
Replacing this with new PR for a new workplan.

PR #690, opened 2022-03-17
Single packing

shawkins:
This is from the investigation of MGDSTRM-7836.  This proposes a set of sizings for our pods such that we can fit 10 managedkafkas per 1 m5.2xlarge node.  To enable this set of values, you have to override the QUARKUS_PROFILE in the FSO deployment to trial.

At .8 MB/s of message data most tests run by the instance profiler (which are at or exceeding the capacity values currently proposed) the performance is more than acceptable.  The cpu does tend to hover at the limit, so we are over committing the cpu by a little - if all 10 brokers on same node were running at .8 MB/s or greater, you'd expect to see increased latency.  As noted on the issue though you see increasing latency after that point anyway due to the affect of the quota.

We will late address support of multiple profiles at a time.


rareddy (changes requested):

rareddy:
> you have to override the QUARKUS_PROFILE in the FSO deployment to trial.

I am wondering how this can be done? This is an operator-level property, so this needs to be set when we are installing FSO, so we can not add it into the SOP for OSD creation. since we do not have different MT pipelines for different OSD types, we can not modify the MT at bundle deployment time either. We can manually set it, but I prefer to remove the human element there.

operator/src/main/resources/application-trial.properties
```diff
@@ -0,0 +1,31 @@
...
+
+managedkafka.canary.container-cpu=75m
+managedkafka.canary.container-memory=64Mi
+
+managedkafka.adminserver.container-request-cpu=75m
+managedkafka.adminserver.container-cpu=200m
+managedkafka.adminserver.container-memory=256Mi
+
+managedkafka.zookeeper.jvm-xms=512M
+managedkafka.zookeeper.container-reqeust-cpu=75m
```
shawkins, line 21:
```suggestion
managedkafka.zookeeper.container-request-cpu=75m
```

shawkins:
> I am wondering how this can be done?

If we lack a mechanism to change the deployment ahead of time, the next option is to use the kas-fleetshard-config configmap with these properties instead.

MikeEdgar:
> > I am wondering how this can be done?
> 
> If we lack a mechanism to change the deployment ahead of time, the next option is to use the kas-fleetshard-config configmap with these properties instead.

Can the addon secret be used to set the profile? I think it is loaded very early in the startup process so perhaps it's possible?

shawkins:
> Can the addon secret be used to set the profile? I think it is loaded very early in the startup process so perhaps it's possible?

That's currently only for sync. I'm honestly not sure if that is earlier enough, but I can test it out.  At worst we'd have to add that same optional handling that you are adding to sync.  The logic to use kas-fleetshard-config exists, so I'll try it there first (just setting the profile, rather than all of the properties).

rareddy:
There is a comment on this in the epic brief, lets's confirm there what we would like there with the control plane team before we commit to the code here IMO.

shawkins:
> There is a comment on this in the epic brief, lets's confirm there what we would like there with the control plane team before we commit to the code here IMO.

I've updated this with changes to the instance profiler as well.  

It's fine to commit these changes for the cleanup to the configuration handling even if we have to go a different direction with the usage of a configmap.

rareddy (approved):
lgtm

shawkins:
Merged to begin work on letting the selection of the profile be driven by something (TBD) in the managedkafka - that will be a much larger change.

PR #692, opened 2022-03-18
Remove strimzi duplicated tests

kornys:
Because this functionality/behavior is daily tested in scope of strimzi tests we really do not need to test in fleetshard operator tests.


MikeEdgar (commented):
Doesn't having these tests still have the benefit of testing the process one level higher in the stack, plus deploying the other pieces like canary and admin-server? Or is your position that it's tested elsewhere in systemtests so these are redundant?

kornys:
In scope of upgrade/smoke test we test deployment of canary/admin. When I checked strimzi suite last time I saw that they already test ability to recover kafka/pods etc so I think these two tests are redundant in fleetshard.

MikeEdgar (approved):

PR #696, opened 2022-03-23
Configure default sync properties for standalone deployment

MikeEdgar:
Allows sync to run (without crash-loops) in environments without external SSO or configuration secret being available.

With this change, the value of env `SSO_ENABLED` must be set to `true` to re-enable the functionality turned off by default.


shawkins (approved):
LGTM

rareddy (approved):
lgtm

PR #697, opened 2022-03-24
MGDSTRM-7817 allow for the configuration of multiple profiles

shawkins:
This is an initial draft of supporting multiple profiles at runtime. Since we need access to all of the properties at runtime, we cannot  use a quarkus profile based mechanism.

- any property that may change based upon the instance type must be moved onto the instance config.
- it treats the existing managedkafka. properties as the standard/default
- additional well-known instances profiles are built off of that looking for properties that are prefixed by the profile name
- it does not yet look for where the instance type is defined

Possible refinements:
- it was basically search replace on the usage of the config, a lot of those calls could be consolidated
- the key names are getting pretty long.  We could change from managedkafka. to standard., and from trial.managedkafka, to just trial.
- using the standard profile as the default means that we need to be careful about changing those values.  If not explicitly overridden, it will change in all instance types.  The alternative is to introduce separate default and standard property sets.
- this does not yet account for different machine types (m5.2xlarge vs m5.4xlarge) or different providers (aws vs azure).

This of course also breaks logic I just added with the last pr for the instranceprofiler, but it will be easy to fix in subsequent commits.


rareddy:
ack.
> key names are getting pretty long

maybe could move to YAML format, you load into that anyways.

>  The alternative is to introduce separate default and standard property sets.

separate one seems less confusing IMO

> this does not yet account for different machine types

essentially these become your keys to fetch the config like STANDARD now. YAML format may ease the pain in defining them.

shawkins:
> maybe could move to YAML format, you load into that anyways.

Sorry, yes forgot to mention that will help / force the config file(s) to be more organized.

> essentially these become your keys to fetch the config like STANDARD now. YAML format may ease the pain in defining them.

My thinking is that I'll move back to separate files, regardless of the file type.  They would initially have a naming standard like instance-platform - for example:

default-aws.yaml
trial-aws.yaml
standard-aws.yaml
and so forth

Then if needed we'll expand that to include the machine type instance-platform-machine - with the caveat that if we want to get really complex you could have standard offered under the same cluster on different machine types in which case we'd need other selection criteria to determine the placement.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@Startup
+@ApplicationScoped
+public class KafkaInstanceConfigurations {
+
+    public enum Config {
+        STANDARD,
+        TRIAL
```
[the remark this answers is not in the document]
rareddy: I was under the impression we wanted call trial everywhere, but this is OK with me if we want to take developer nomenclature in code.

rareddy:
> My thinking is that I'll move back to separate files, regardless of the file type.

sounds good me.

shawkins:
> Sorry, yes forgot to mention that will help / force the config file(s) to be more organized.

This is possible to do, but requires changing the kafka instance configuration to not use jsonunwrap and we'd have to change the logic about how we infer the property key set.  So for now I've just left things as property files.

Since the non application.properties properties files are managed by us and not quarkus, we can omit the managedkafka or any other prefix in them.

If you need to override them from an env property or other quarkus supported mechanism they do need to be prefixed - developer.kafka.something.

operator/src/main/resources/application.properties
```diff
@@ -129,50 +136,16 @@ managedkafka.zookeeper.volume-size=10Gi
 managedkafka.kafka.max-connections=3000
```
rareddy, line 136: IMO we should totally remove any properties represented by KafkaInstaceConfiguration in default `application.properties`. It may get confusing what default, who overridden what etc.

shawkins: Without additional refactoring some have to remain as they aren't actually part of the KafkaInstanceConfiguration.  I'm fine with leaving the rest of the the defaults in the main application.properties.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,93 @@
...
+ * <p>
+ * Properties in each instance file will omit this prefix.
+ * <p>
+ * Properties overlayed from the Quarkus Config will be prefixed accordingly - managedkafka.kafka... standard.kafka...
+ */
+@Startup
+@ApplicationScoped
+public class KafkaInstanceConfigurations {
+
+    private static final String MANAGEDKAFKA = "managedkafka";
```
rareddy, line 31: why not remove this prefix everywhere, or since the property files are already named differently have the same prefix on all?

shawkins: > why not remove this prefix everywhere

The original intent of the managedkafka prefix was to organize our own properties in the application.properties file.  Stuff with that prefix would be different than properties that were for configuring internal things like polling intervals.

We also need to make sure that any changes to the property prefix, or even what I've done to introduce a different standard set of property doesn't conflict with anything that is already part of an SOP.

> or since the property files are already named differently have the same prefix on all

You could have the same, but I prefer having no prefix in those files than having managedkafka. - because if you were to actually override them it would need to be standard. or trial. etc.

operator/src/main/resources/application.properties
shawkins: Several of the canary ones at least.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
rareddy: seems like instead of depending on `application.properties` if you introduce a `default.properties` then with prefix one place vs no prefix other will fix itself?

operator/src/main/resources/application.properties
rareddy: If are saying these
```
    @ConfigProperty(name = "managedkafka.canary.producer-latency-buckets")
    @ConfigProperty(name = "managedkafka.canary.endtoend-latency-buckets")
    @ConfigProperty(name = "managedkafka.canary.connection-check-latency-buckets")
    @ConfigProperty(name = "managedkafka.canary.status-time-window-ms")
    @ConfigProperty(name = "managedkafka.canary.init-timeout-seconds")
    @ConfigProperty(name = "managedkafka.canary.init-enabled")
```
seems like they need to be fixed to be consistent imo.

shawkins: My point is that it's optional work - not tied to what is being asked for here.  Only if you think those properties will get changed per instance type do we to have them at this time on the kafkainstanceconfiguration.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
shawkins: You could along with the change to move all such properties on to the kafkainstanceconfiguration, but again it didn't seem necessary at this juncture.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public enum Config {
+        STANDARD,
+        DEVELOPER
+    }
```
MikeEdgar, line 36: Should this be something like `InstanceType` rather than `Config`? It's cosmetic, but at least for me it looked like a `org.eclipse.microprofile.config.Config` instance was being passed around in this class.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private void loadConfiguration(Map<String, String> defaultValues, Config toLoad) throws IOException {
+        String name = toLoad.name().toLowerCase();
+        Map<String, String> allValues = new HashMap<>(defaultValues);
+
+        Properties p = new Properties();
+        try (InputStream stream = this.getClass().getResourceAsStream(String.format("/%s.properties", name))) {
+            p.load(stream);
+        }
```
MikeEdgar, line 58: Consider adding some conventional prefix to the name of the properties files.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,93 @@
...
+
+        // set the properties from the file
+        allValues.keySet().forEach(n -> {
+            String key = n.substring(MANAGEDKAFKA.length() + 1);
+            String value = p.getProperty(key);
+            if (value != null) {
+                allValues.put(n, value);
+            }
```
MikeEdgar, line 66: I was thinking there might be a risk of `ConcurrentModificationException` here, but there isn't because it's only changing the value of an existing entry. It might be better/clearer to iterate over the entry set and use `Entry.setValue`.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,93 @@
...
+                Serialization.jsonMapper().convertValue(allValues, KafkaInstanceConfiguration.class);
+        configs.put(toLoad, config);
+    }
+
+    private void overlayWithConfigProperties(Map<String, String> defaultValues, UnaryOperator<String> keyMapper) {
+        // overlay anything from the quarkus config
+        defaultValues.keySet()
+                .forEach(n -> ConfigProvider.getConfig()
+                        .getOptionalValue(keyMapper.apply(n), String.class)
+                        .ifPresent(v -> defaultValues.put(n, v)));
```
MikeEdgar, line 81: Same comment above regarding `entrySet` iteration.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
MikeEdgar: I tend to agree that having the configs for this out of `application.properties` is easier for reference purposes. But, I also like the option to override using the environment or system properties.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
shawkins: Right there's no risk of a CME as those will only happen from structural modifications.  No problem on changing it to use the entries.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+@Startup
+@ApplicationScoped
+public class KafkaInstanceConfigurations {
+
+    private static final String MANAGEDKAFKA = "managedkafka";
+
+    public enum Config {
+        STANDARD,
+        DEVELOPER
```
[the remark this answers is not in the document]
shawkins: Yes, that will be fine.  It's just that the convention for java is upper case enum names.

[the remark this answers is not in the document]
machi1990: Perfect, thanks for confirming. 

Yes, I know. And some json libraries tend to rely on the enum name not ignoring the case.

shawkins:
> I tend to agree that having the configs for this out of application.properties is easier for reference purposes. But, I also like the option to override using the environment or system properties.

They will still be overridable via standard quarkus mechanisms.

operator/src/main/resources/application.properties
rareddy: agree, they are not related. I was looking from the point of view to keep it consistent with other properties. the way these properties are defined the author seemed to think they can change at the instance level but did not take the time to do it consistently with others IMO. we can come back in another JIRA and fix them then.

MikeEdgar (approved):
LGTM :+1: 

> The only way good way address to address that last one is to redo all of this using https://quarkus.io/guides/config-extending-support#config-source-factory (or similar mechanism such that when we get the property value from the Config substitution has been performed).

I agree this could be dealt with separate from this PR if necessary.

rareddy (approved):
lgtm

PR #698, opened 2022-03-24
Cruisecontrol integration

k-wall:
Enables Cruise Control support.

### DIFF
--- a/operator/src/main/resources/instances/developer.properties
+++ b/operator/src/main/resources/instances/developer.properties
@@ -26,3 +26,5 @@ storage.min-margin=1Gi
 # this is arbitrary, we don't expect request for multiple developer units
 kafka.partition-capacity=1500
 kafka.message-max-bytes=1048588
+
+cruisecontrol.enabled=false
--- a/operator/src/main/resources/instances/managedkafka.properties
+++ b/operator/src/main/resources/instances/managedkafka.properties
@@ -31,3 +31,15 @@ storage.formatting-overhead=0.05
 
 # Override for partition limit enforcement. This override should be removed once we can enable this by default
 kafka.partition-limit-enforced=false
+
+cruisecontrol.metric-reporter-topic=__redhat_strimzi.cruisecontrol.metrics
+cruisecontrol.metric-reporter-topic-min-insync-replicas=2
+
+cruisecontrol.partition-metric-sample-store-topic=__redhat_strimzi.cruisecontrol.partitionmetricsamples
+cruisecontrol.broker-metric-sample-store-topic=__redhat_strimzi.cruisecontrol.modeltrainingsamples
+cruisecontrol.metric-sample-store-topic-replication-factor=3
+
+cruisecontrol.hard-goals=com.linkedin.kafka.cruisecontrol.analyzer.goals.RackAwareGoal,com.linkedin.kafka.cruisecontrol.analyzer.goals.DiskUsageDistributionGoal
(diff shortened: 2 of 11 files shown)


operator/src/main/resources/instances/developer.properties
```diff
@@ -21,7 +21,7 @@ kafka.container-memory=1Gi
 kafka.jvm-xms=512M
 
 cruisecontrol.container-cpu=75m
-cruisecontrol.container-memory=256Mi
+cruisecontrol.container-memory=1Gi
```
shawkins, line 24: @k-wall @katheris since there is only a single broker will cruisecontrol even need to be enabled with developer instances?

katheris: @shawkins that is a good point. @k-wall do you think we would ever enable cruise control for the developer instances? It feels like a production feature to me, and as Steven pointed out it can't do much in a single broker cluster anyway

k-wall: until we support multi volumes per broker, I don't think there is any value in CC for developer.

operator/src/main/resources/instances/standard.properties
```diff
@@ -13,6 +13,10 @@ zookeeper.jvm-xms=1G
 zookeeper.container-cpu=1000m
 zookeeper.container-memory=4Gi
 
+cruisecontrol.container-memory=2Gi
```
shawkins, line 16: Between this and the 4.10 resource reduction (there's 336Mi less allocatable per node) we'll make an adjustment to the kafka.container-memory.  Assuming that we want to leave 1Gi of padding, the new number would be 17.3Gi instead of 18.1Gi.

With the cpu this low, I don't believe that we need to make any adjustment to the broker cpu.

shawkins: @rareddy @katheris @k-wall This is assuming that cruise control could be enabled even with the 3 brokers.  If it will always remain 6 or more, then we don't need to make any resource changes to account for cruise control.

rareddy: I suspect we eventually come around to the model where CC will be turned on with the 3 brokers, so maybe IMO it is better to align to that to begin with.

operator/src/main/resources/instances/managedkafka.properties
```diff
@@ -31,3 +31,15 @@ storage.formatting-overhead=0.05
 
 # Override for partition limit enforcement. This override should be removed once we can enable this by default
 kafka.partition-limit-enforced=false
+
+cruisecontrol.metric-reporter-topic=__redhat_strimzi.cruisecontrol.metrics
```
rareddy, line 35: I remember in another issue where we are renaming from "_redhat" to "_kas"?

katheris: @rareddy are you able to find that issue? I've had a look but can't find it, and don't want to rename unless we've changed the internal logic that determines which topics should be hidden from customers

katheris:
@shawkins I've pushed an additional change to update the Kafka memory as you suggested. Can you re-review this PR?

shawkins (approved):
LGTM

operator/src/main/resources/instances/managedkafka.properties
k-wall: I think __redhat is okay here for now.  I think there is a separate task to de-Red Hat kas-fleetshard as part of its open-source story.  We can tackle all the __redhat references in one go.

rareddy: sounds good (y)

PR #699, opened 2022-03-28
[WIP] Factor out inline secret removal

biswassri:
[MGDSTRM-5205](https://issues.redhat.com/browse/MGDSTRM-5205) 

bootstrap certificates
oauth - clientid, clientSecret
Canary service account

[Replacing #687 ]


sync/src/main/java/org/bf2/sync/informer/InformerManager.java
```diff
@@ -46,6 +51,19 @@ protected void onStart() {
         managedAgentInformer = resourceInformerFactory.create(ManagedKafkaAgent.class, client.resources(ManagedKafkaAgent.class).inAnyNamespace(),
                 CustomResourceEventHandler.of(controlPlane::updateAgentStatus));
 
+        secretInformer = resourceInformerFactory.create(Secret.class, client.secrets().inAnyNamespace().withLabels(OperandUtils.getDefaultLabels()),
```
MikeEdgar, line 54: I wonder if there should be an additional label on the master secret and this informer scoped to only those secrets.

api/src/main/java/org/bf2/operator/resources/v1alpha1/TlsKeyPair.java
```diff
@@ -19,24 +18,48 @@
...
     private String key;
+    private SecretKeySelector certRef;
+    private SecretKeySelector keyRef;
 
     public String getCert() {
-        return cert;
+        if(cert != null){
+            return cert;
+        }
+        return certRef.getKey();
```
MikeEdgar, line 32: I don't think this method should change, continue returning `cert` unconditionally.

api/src/main/java/org/bf2/operator/resources/v1alpha1/TlsKeyPair.java
```diff
@@ -19,24 +18,48 @@
...
     public void setCert(String cert) {
         this.cert = cert;
     }
 
     public String getKey() {
-        return key;
+        if(key != null){
+            return key;
+        }
+        return keyRef.getKey();
```
MikeEdgar, line 43: Same as `getCert`, no change here.

MikeEdgar (commented):
I think this is headed in the right direction. I added some comments on a few things that should be addressed as you continue development.

api/src/main/java/org/bf2/operator/resources/v1alpha1/TlsKeyPair.java
biswassri: So the idea of returning `certRef.getKey()` is to not change with the secret retrieval process in Operator. Wherever we are trying `TlsKeyPair.getCert()` it would continue with the same. But do you think we should change the TlsKeyPair retrieval process everywhere? ( I mean now we do `getCertRef` instead `getCert `?)

MikeEdgar: I don't think that would work. The operator code still needs to change to either copy the secrets to their current locations using the `SecretKeySelector`s or use the refs directly.

sync/src/main/java/org/bf2/sync/informer/InformerManager.java
rareddy: yes, that makes sense, while creating the secret can give additional label to filter on

api/src/main/java/org/bf2/operator/resources/v1alpha1/SecretKeySelector.java
```diff
@@ -0,0 +1,88 @@
...
+public class SecretKeySelector implements KubernetesResource
+{
+
+    @JsonProperty("key")
+    private String key;
+    @JsonProperty("name")
+    private String name;
+    @JsonProperty("optional")
+    private Boolean optional;
+
```
rareddy, line 35: what is the issue with `addtionalProperties` here that we had to clone this class?

sync/src/main/java/org/bf2/sync/SecretManager.java
```diff
@@ -0,0 +1,231 @@
...
+
+    public static boolean isKafkaExternalCertificateEnabled(ManagedKafka managedKafka) {
+        return managedKafka.getSpec().getEndpoint().getTls() != null;
+    }
+
+    public static boolean isCanaryServiceAccountPresent(ManagedKafka managedKafka){
+        return managedKafka.getServiceAccount(ServiceAccount.ServiceAccountName.Canary).isPresent();
+    }
+
+    public boolean masterSecretExists(ManagedKafka managedKafka){
```
rareddy, line 56: maybe this should be changed to `isMasterSecretExists` to meet the other methods nomenclature

sync/src/main/java/org/bf2/sync/SecretManager.java
```diff
@@ -0,0 +1,231 @@
...
+        return cachedOrRemoteSecret(managedKafka,masterSecretName(managedKafka)) != null;
+    }
+
+    public boolean masterSecretChanged(ManagedKafka remote, ManagedKafka local){
```
rareddy, line 60: maybe this should be changed to `ismasterSecretChanged` to meet the other methods nomenclature

sync/src/main/java/org/bf2/sync/SecretManager.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private Map<String, String> ssoClientDataFrom(ManagedKafka managedKafka) {
+        return Map.of(OAUTH_SSO_CLIENT_ID, managedKafka.getSpec().getOauth().getClientId(),
+                OAUTH_SSO_CLIENT_SECRET, managedKafka.getSpec().getOauth().getClientSecret());
+    }
+
+    private Map<String, String > canarySaslDataFrom(ManagedKafka managedKafka){
+        return Map.of(CANARY_SASL_PRINCIPAL, managedKafka.getServiceAccount(ServiceAccount.ServiceAccountName.Canary).get().getPrincipal(),
+                CANARY_SASL_PASSWORD, managedKafka.getServiceAccount(ServiceAccount.ServiceAccountName.Canary).get().getPassword());
+    }
```
rareddy, line 230: maybe these are better inlined into the method where they are called?

api/src/main/java/org/bf2/operator/resources/v1alpha1/SecretKeySelector.java
MikeEdgar: It's annotated with `@JsonIgnore` which has a `value` attribute that's a `boolean`. The version of fabric8 we're using reads `value` generically and casts it to a string without checking the type first.

operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
```diff
@@ -213,21 +221,40 @@ private static Secret buildSecretFrom(String name, String type, ManagedKafka man
...
+        }
+        else {
+            return buildSecretFrom(kafkaTlsSecretName(managedKafka),
+                    "kubernetes.io/tls",
+                    managedKafka,
+                    current,
+                    Map.of("tls.crt", managedKafka.getSpec().getEndpoint().getTls().getCert(),
+                            "tls.key", managedKafka.getSpec().getEndpoint().getTls().getKey()));
+        }
     }
```
MikeEdgar, line 241: Only a suggestion, but I think it's easier to read if only the parts that depend on the condition are in the if/then/else blocks.
```suggestion
        String cert;
        String key;
        
        if (currentMasterSecret != null) {
            cert = currentMasterSecret.getData().get("endpoint.tls.crt");
            key = currentMasterSecret.getData().get("endpoint.tls.key");
        } else {
            cert = managedKafka.getSpec().getEndpoint().getTls().getCert();
            key = managedKafka.getSpec().getEndpoint().getTls().getKey();
        }

        return buildSecretFrom(kafkaTlsSecretName(managedKafka),
                "kubernetes.io/tls",
                managedKafka,
                current,
                Map.of("tls.crt", cert,
                       "tls.key", key));
    }
```

MikeEdgar:
@biswassri, I added a commit with what I think will work in the sync. The idea is to keep the check of the secret digest more closely related to the spec comparison and to separate the digest calculation from the secret replacement code. Let me know what you think.

operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
biswassri: Sure! I could change the if/then/else blocks accordingly.

sync/src/main/java/org/bf2/sync/SecretManager.java
```diff
@@ -108,67 +100,67 @@ private Secret buildSecret(String name, String type, ManagedKafka managedKafka,
...
                 .build();
-        return managedKafka;
+    }
+
+    public void calculateMasterSecretDigest(ManagedKafka managedKafka, Secret secret) {
+        var metadata = new ObjectMetaBuilder(managedKafka.getMetadata())
+                .addToAnnotations(ANNOTATION_MASTER_SECRET_DIGEST, buildDigest(decode(secret.getData())))
+                .build();
+
+        managedKafka.setMetadata(metadata);
```
biswassri, line 163: @MikeEdgar Could you help with a little bit more information about why we creating this `calculateMasterSecretDigest` method and then building the digest?

biswassri:
Replaced by https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/712

PR #700, opened 2022-03-31
Make OAuth clientId and clientSecret optional

MikeEdgar:
MGDSTRM-7968


MikeEdgar:
@biswassri FYI this impacts the changes you are making to remove the secrets from the ManagedKafka resource.

biswassri:
> Yes, with your PR (once this is merged in) it will need to check for the presence of the secret references since the secret value will always be null in the operator.

Right! Makes sense. I'll include that as well.

MikeEdgar:
> For the existing CRs, will it remove the configured clientid/secret? or it will only impact the new Kafka request?

This change only allows the clientid/secret to be unspecified in the CR. Existing CRs would need to have them removed in the control plane.

akoserwal:
@MikeEdgar thanks

vbusch (approved):
lgtm

PR #702, opened 2022-03-31
MGDSTRM-7965 using and propagating instance/profile type

shawkins:
This is an initial draft of using the profile type from the control plane.

- it does not use the quota directly as we haven't reached any kind of consensus on that
- makes a minor change so that delete can provide an appropriate status update even if invalid
- adds in a speculative scheduling change for when we mix deployments - the zookeepers have an anti-affinity to anything that is not the same profile.  This will need validated and potentially relaxed in the future.  Beyond that we're just going to rely on the autoscaling to compact things for us.


rareddy:
1) IMO, delete stuff should be in a separate PR
2) as per anti-affinity stuff, my preference is still to follow ADR-68 to have different machine pools to place the instances, could we defer this to later?

shawkins:
> IMO, delete stuff should be in a separate PR

It can be, it's just related to the overall restructuring (it's an additional one line change https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/702/files#diff-1c7d44788e015d2397ec69318d7e7b5c9ddf960063aab5e7d3aa2180ba286b81R156) in the status logic in the managedkafkacontroller. 

> as per anti-affinity stuff, my preference is still to follow ADR-68 to have different machine pools to place the instances, could we defer this to later?

We can - it's intentionally speculative and not meant to be merged.  My thinking is that ADR-68 is less necessary within just the managed kafka due to auto-scaling - as my primary concern had been the compaction logic.  If we expand this type of scheduling beyond the zookeeper pods in a similar manner, then we can share a single machine pool for standard and developer.  ADR-68 is still conceptually needed for segregating for from different managed services.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -77,12 +87,15 @@ private void overlayWithConfigProperties(Map<String, String> defaultValues, Unar
...
     }
 
     public KafkaInstanceConfiguration getConfig(ManagedKafka managedKafka) {
-        // TDB where to get name from
-        return configs.get(InstanceType.STANDARD);
+        return configs.get(getInstanceType(managedKafka));
+    }
+
+    public static String getInstanceType(ManagedKafka managedKafka) {
+        return OperandUtils.getOrDefault(managedKafka.getMetadata().getLabels(), PROFILE_TYPE, InstanceType.STANDARD.lowerName);
```
MikeEdgar, line 98: It might be useful for development to allow a config property to override the instance type. For example
```
mvn quarkus:dev -Dmanagedkafka.instancetype-override=DEVELOPER
```

shawkins: I'm not sure when we would use that.  If we're already in the dev profile we're not creating real instances.  You can also just declaratively update the managedkafka to be a developer instance.

shawkins:
The next commit adds the logic to sync to detect metadata changes as well.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
MikeEdgar: > If we're already in the dev profile we're not creating real instances.

What do you mean by that? I frequently run it in dev mode to create/update instances in OSD. Prior to profiles I would override the resource requests manually, but setting this to `DEVELOPER` now that there are profiles is much more convenient.

shawkins: > What do you mean by that?

In quarkus:dev the dev operands are used - see the package org.bf2.operator.operands.dev - these are not the same as standard or developer instances.

shawkins: > but they're close enough for a lot of scenarios

Sure it's primarily for minikube.  It's unfortunate that we have yet another overloaded term :(  To recap - in dev mode you won't even hit this logic in KafkaCluster related to profiles and configuration overrides.  You just get the dev instance.  If for whatever reason you want the managedkafka to have the profile label, it can be there, but won't get used or propagated.

operator/src/test/resources/expected/developer-strimzi.yml
```diff
@@ -0,0 +1,295 @@
+---
+kind: "Kafka"
+metadata:
+  annotations:
+    org.bf2.operator/desired-broker-replicas: "1"
+  labels:
+    app.kubernetes.io/managed-by: "kas-fleetshard-operator"
+    ingressType: "sharded"
+    managedkafka.bf2.org/strimziVersion: "strimzi-cluster-operator.v0.23.0-4"
+    bf2.org/kafkaInstanceProfileType: "developer"
```
rareddy, line 10: I did not notice before but should we negotiate `bf2.org` to be `managedkafka.bf2.org` here for consistency?

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
```diff
@@ -44,7 +51,7 @@ void init() throws IOException {
         loadConfiguration(defaultValues, MANAGEDKAFKA);
 
         for (InstanceType type : InstanceType.values()) {
-            configs.put(type, loadConfiguration(new HashMap<>(defaultValues), type.name().toLowerCase()));
+            configs.put(type.name().toLowerCase(), loadConfiguration(new HashMap<>(defaultValues), type.name().toLowerCase()));
```
rareddy, line 54: multiple types of usage of this lower cased name I see, maybe better to pull all these into single method?

grdryn:
Please add the label to canary and admin server deployments (both to the `metadata.labels` and to `spec.template.metadata.labels` to ensure that it gets propagated down to Pods. This will be needed for the changes I'm making in https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/pull/177.

shawkins:
sure thing, I'll add it to all of the other pods.

operator/src/test/resources/expected/developer-strimzi.yml
grdryn: If changing it, please let me know, so that I can update https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/pull/177 to match.

I was going to suggest also changing to `kafka-instance-profile-type`, but I see there isn't consistency in the already existing labels, so probably not worth worrying about.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfigurations.java
shawkins: Sorry, hadn't switched those over to lowerName yet.

operator/src/test/resources/expected/developer-strimzi.yml
shawkins: @rareddy so you are looking for managedkafka.bf2.org/instance-profile-type?

I don't have a strong preference, but I would address consistency beyond these new labels with another issue.  Any label changes we'd want to make will likely just need to duplicate the labels for a while.

rareddy: I was only suggesting these ones for now as these are new labels, why go through another sync point across Control Plane, Kafka Integrations and Fleet Services team later. if we want address it, do it now or leave it be IMO. 

@pb82 FYI..

grdryn:
@shawkins FYI this needs a rebase (conflicts).

shawkins:
@grdryn resolved now.

operator/src/main/java/org/bf2/operator/operands/Canary.java
```diff
@@ -260,8 +260,9 @@ private Map<String, String> buildSelectorLabels(String canaryName) {
         return labels;
     }
 
-    private Map<String, String> buildLabels(String canaryName) {
+    Map<String, String> buildLabels(String canaryName, ManagedKafka managedKafka) {
         Map<String, String> labels = buildSelectorLabels(canaryName);
+        Optional.ofNullable(managedKafka.getMetadata().getLabels()).ifPresent(labels::putAll);
```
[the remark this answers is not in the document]
shawkins: If managedKafka.getMetadata().getLabels() is null or empty, no labels will be added to the "labels" map.

MikeEdgar (approved):
LGTM

operator/src/main/java/org/bf2/operator/operands/Canary.java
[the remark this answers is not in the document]
rareddy: ok, leaving it empty seems appropriate. thanks

rareddy (approved):
lgtm

machi1990 (approved):
checked the labels against what Control Plane sends and they look good to me

PR #703, opened 2022-04-06
Configure the soft and hard limits at customer's provisioned capacity.

SamBarker:
We don't believe customers will notice producer throttling in time to act, a safety factor of 2 there would have been a little over 1 min of throttling before the hard limit. Moving to a factor of 5 would give them 2m 30s or so to react.
The hard limit is what really protects the broker anyway, so we just simplify the documentation and monitoring story to the hard limit.

There is an outstanding question should the hard limit be at or slightly beyond the provisioned storage?

https://issues.redhat.com/browse/MGDSTRM-8110

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -607,14 +607,15 @@ public class KafkaCluster extends AbstractKafkaCluster {
         config.put(QUOTA_FETCH, String.valueOf(getEgressBytes(managedKafka, current)));
 
         // Start throttling when disk is above requested size. Full stop when only storageMinMargin is free.
-        Quantity maxDataRetentionSize = getAdjustedMaxDataRetentionSize(managedKafka, current);
-        KafkaInstanceConfiguration instanceConfig = this.configs.getConfig(managedKafka);
-        long hardStorageLimit = Quantity.getAmountInBytes(maxDataRetentionSize).longValue() - Quantity.getAmountInBytes(instanceConfig.getStorage().getMinMargin()).longValue();
-        long softStorageLimit = Quantity.getAmountInBytes(maxDataRetentionSize).longValue() - getStoragePadding(managedKafka, current);
-        config.put("client.quota.callback.static.storage.soft", String.valueOf(softStorageLimit));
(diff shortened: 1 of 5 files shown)


k-wall (approved):

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -607,14 +607,15 @@ private void addQuotaConfig(ManagedKafka managedKafka, Kafka current, Map<String
...
-        Quantity maxDataRetentionSize = getAdjustedMaxDataRetentionSize(managedKafka, current);
-        KafkaInstanceConfiguration instanceConfig = this.configs.getConfig(managedKafka);
-        long hardStorageLimit = Quantity.getAmountInBytes(maxDataRetentionSize).longValue() - Quantity.getAmountInBytes(instanceConfig.getStorage().getMinMargin()).longValue();
-        long softStorageLimit = Quantity.getAmountInBytes(maxDataRetentionSize).longValue() - getStoragePadding(managedKafka, current);
-        config.put("client.quota.callback.static.storage.soft", String.valueOf(softStorageLimit));
-        config.put("client.quota.callback.static.storage.hard", String.valueOf(hardStorageLimit));
+        final long maxRetentionBytes = Quantity.getAmountInBytes(getAdjustedMaxDataRetentionSize(managedKafka, current)).longValue();
+        final long storagePaddingBytes = getStoragePadding(managedKafka, current);
+        long storageLimit = maxRetentionBytes - storagePaddingBytes;
+        //For RHOASK its unlikely that customers will notice producer throttling, so we set it to the same as the hard limit and let the hard limit win
```
k-wall, line 613: We shouldn't mention this name in the code, even mispelt

shawkins (approved):
LGTM, added a comment to https://docs.google.com/document/d/1JGyj_z2G61Ept3gg1P0f26vAd1UpEOmbpTcNFkSYRus/edit to make the doc effort around the capacity values aware of this change.

SamBarker:
> LGTM, added a comment to https://docs.google.com/document/d/1JGyj_z2G61Ept3gg1P0f26vAd1UpEOmbpTcNFkSYRus/edit to make the doc effort around the capacity values aware of this change.

Thanks Steve, I was wondering where to flag it for docs (not that I have access to that doc but I'm sure that will be sorted soon).

PR #704, opened 2022-04-07
bugfix: compensate for the ext4 formatting overhead in storage calculations

k-wall:
Previously the code assumed that the PV capacity's was the same same as the available space on the volume provisioned by OpenShift CSI.  This fails to account for the formatting overhead.

The service currently relies on the default `storageclass`, which uses a default `fsType`, which defaults to `ext4`. 

Actually calculating the exact formatting overhead for ext4 is non-trivial.  Instead we use a 5% heuristic which is shown to be sufficient for storage sizes of 16Ti and below (the ext4 maximum).  Note that volumes are formatted with the `reserved for root` disabled, so we don't need to consider that.

When this code is deployed, any existing kafka instances will have their storage increased by 5%.

There's more discussion on https://issues.redhat.com/browse/MGDSTRM-8072

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -195,8 +195,6 @@ public class KafkaCluster extends AbstractKafkaCluster {
         int desiredReplicas = getBrokerReplicas(managedKafka, null);
 
         long storagePerBroker = getPerBrokerBytes(managedKafka.getSpec().getCapacity().getMaxDataRetentionSize(), () -> this.configs.getConfig(managedKafka).getKafka().getVolumeSize(), actualReplicas);
-        long safetyMargin = calculateSafetyMargin(managedKafka, current);
-        long formatOverheadWithMargin = safetyMargin + calculateFormattingOverhead(managedKafka, storagePerBroker + safetyMargin, FormattedValueType.FORMATTED);
 
         KafkaInstanceConfiguration config = this.configs.getConfig(managedKafka);
         KafkaBuilder kafkaBuilder = builder
@@ -214,7 +212,7 @@ public class KafkaCluster extends AbstractKafkaCluster {
                         .withReplicas(actualReplicas)
                         .withResources(config.kafka.buildResources())
                         .withJvmOptions(buildKafkaJvmOptions(managedKafka))
-                        .withStorage(buildKafkaStorage(managedKafka, current, storagePerBroker, formatOverheadWithMargin))
+                        .withStorage(buildKafkaStorage(managedKafka, current, storagePerBroker))
                         .withListeners(buildListeners(managedKafka, actualReplicas))
                         .withRack(buildKafkaRack(managedKafka))
                         .withTemplate(buildKafkaTemplate(managedKafka))
@@ -625,10 +623,14 @@ public class KafkaCluster extends AbstractKafkaCluster {
         config.put("quota.window.size.seconds", "2");
     }
 
-    private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current, long storagePerBroker, long storagePadding) {
+    private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current, long storagePerBroker) {
+        long storageBytes = storagePerBroker;
+        storageBytes += calculateSafetyMargin(managedKafka, current);
+        storageBytes += calculateFormatOverheadFromFormattedSize(managedKafka, storageBytes);
+
         PersistentClaimStorageBuilder builder = new PersistentClaimStorageBuilder()
                 .withId(JBOD_VOLUME_ID)
-                .withSize(getAdjustedMaxDataRetentionSize(current, storagePerBroker, storagePadding).getAmount())
+                .withSize(getAdjustedMaxDataRetentionSize(current, storageBytes).getAmount())
                 .withDeleteClaim(DELETE_CLAIM);
 
         Optional.ofNullable(current).map(Kafka::getSpec).map(KafkaSpec::getKafka).map(KafkaClusterSpec::getStorage)
@@ -690,31 +692,22 @@ public class KafkaCluster extends AbstractKafkaCluster {
                         * config.getStorage().getSafetyFactor();
     }
 
-    enum FormattedValueType {
-        FORMATTED, UNFORMATTED
+    protected long calculateFormatOverheadFromUnformattedSize(ManagedKafka managedKafka, long size) {
+        double formattingOverhead = this.configs.getConfig(managedKafka).getStorage().getFormattingOverhead();
+        return (long) (size - (size / (1 + formattingOverhead)));
     }
 
-    /**
-     * Given either a formatted or unformatted size, calculate the formatting overhead using a heuristic.
-     */
-    protected long calculateFormattingOverhead(ManagedKafka managedKafka, long size, FormattedValueType formattedValueType) {
+    protected long calculateFormatOverheadFromFormattedSize(ManagedKafka managedKafka, long size) {
         double formattingOverhead = this.configs.getConfig(managedKafka).getStorage().getFormattingOverhead();
-        if (formattedValueType == FormattedValueType.FORMATTED) {
-            return (long) (size * formattingOverhead);
-        } else {
-            return (long) (size - (size / (1 + formattingOverhead)));
-        }
+        return (long) (size * formattingOverhead);
     }
 
     /**
      * Get the effective volume size considering extra padding and the existing size
      */
-    private Quantity getAdjustedMaxDataRetentionSize(Kafka current, long storagePerBroker, long storagePadding) {
-        long bytes = storagePerBroker + storagePadding;
-
-        // pad to give a margin before soft/hard limits kick in
-
+    private Quantity getAdjustedMaxDataRetentionSize(Kafka current, long storageBytes) {
         // strimzi won't allow the size to be reduced so scrape the size if possible
+        long bytes = storageBytes;
         if (current != null) {
             Storage storage = current.getSpec().getKafka().getStorage();
             if (storage instanceof JbodStorage) {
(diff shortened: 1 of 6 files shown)


operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
```diff
@@ -60,6 +60,10 @@
         @JsonProperty("min-margin")
         protected Quantity minMargin;
 
+        @JsonProperty("formatting-overhead")
+        protected double formattingOverhead = 105d/100;
```
grdryn, line 64: The default value of `105d/100` here is confusing to me. It evaluates to `1.05`, whereas we override in the properties file to `0.05`. Am I reading it correctly? What's the reason for the big difference?

k-wall: that's a mistake left from a previous version.  My intent is to have no default here and rely on the default from managedkafka.properties.

k-wall: fixed.

grdryn (approved):
Looks good to me :+1:

shawkins (commented):
> When this code is deployed, any existing kafka instances will have their storage increased by 5%.

@rareddy @bibryam just looping in on the increase in costs.

Should we differentiate between new and existing instances?  That is only increase storage for existing, but for new just take this overhead out of the usable space?

k-wall:
> > When this code is deployed, any existing kafka instances will have their storage increased by 5%.
> 
> @rareddy @bibryam just looping in on the increase in costs.
> 
> Should we differentiate between new and existing instances? That is only increase storage for existing, but for new just take this overhead out of the usable space?

The contract at the is the the fleetshard to provide nGi of usable storage to the user.  We currently don't do that - so I see that as a defect. The customer should not need to think about how the storage is provided or how it is formatted.

shawkins:
> The customer should not need to think about how the storage is provided or how it is formatted.

The customer isn't directly paying for this storage based upon what is usable - it would simply be a matter of documenting what the space means on https://docs.google.com/document/d/1JGyj_z2G61Ept3gg1P0f26vAd1UpEOmbpTcNFkSYRus/edit

rareddy:
it is true that our current storage payment model is based on usage than total storage allocated. However, until we have dynamic storage in place, we probably should use this hard limit as the customer is not willing to pay more than this amount without pre-authorization. Since we are cost-conscious, maybe it should be ok to draw the hard limit lines below the advertised sizes, as the run-up to the limit user has not been changed for all the or they are going to be charged for.

bibryam:
I'm not able able to follow everything said here. If we need 5% more for safety, that is fine.
When we say 1TB limit, that is where our soft=hard limit should be. Overhead is our concern and cost that user is not even aware.
Steve now sure what you suggest should be documented. I wasn't expected any of this bugfix/implementation to result in the user docs.

rareddy:
To bluntly put it this will increase our AWS storage charges by 5%

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -207,11 +210,11 @@ public Kafka kafkaFrom(ManagedKafka managedKafka, Kafka current) {
                 .editOrNewSpec()
                     .editOrNewKafka()
                         .withVersion(this.kafkaManager.currentKafkaVersion(managedKafka))
-                        .withConfig(buildKafkaConfig(managedKafka, current))
+                        .withConfig(buildKafkaConfig(managedKafka, current, storagePerBroker))
                         .withReplicas(actualReplicas)
                         .withResources(config.kafka.buildResources())
                         .withJvmOptions(buildKafkaJvmOptions(managedKafka))
-                        .withStorage(buildKafkaStorage(managedKafka, current))
+                        .withStorage(buildKafkaStorage(managedKafka, current, storagePerBroker, formatOverheadWithMargin))
```
SamBarker, line 217: To avoid the downward travelling flag should we calculate the `safetyMargin` and `formatOverheadWithMargin` in `buildKafkaStorage`

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -195,6 +194,10 @@ public Kafka kafkaFrom(ManagedKafka managedKafka, Kafka current) {
         int actualReplicas = getBrokerReplicas(managedKafka, current);
         int desiredReplicas = getBrokerReplicas(managedKafka, null);
 
+        long storagePerBroker = getPerBrokerBytes(managedKafka.getSpec().getCapacity().getMaxDataRetentionSize(), () -> this.configs.getConfig(managedKafka).getKafka().getVolumeSize(), actualReplicas);
```
SamBarker, line 197: I'm curious should we use *actual*  or *desired* replicas here? 

I lean towards desired as it reduces the number of storage size changes if actual moves towards desired incrementally. 

I don't feel this needs addressing before this PR merges.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -666,38 +665,55 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current) {
...
      */
-    private long getStoragePadding(ManagedKafka managedKafka, Kafka current) {
+    private long calculateSafetyMargin(ManagedKafka managedKafka, Kafka current) {
         KafkaInstanceConfiguration config = this.configs.getConfig(managedKafka);
         return Quantity.getAmountInBytes(config.getStorage().getMinMargin()).longValue()
                 + getIngressBytes(managedKafka, current) * config.getStorage().getCheckInterval()
                         * config.getStorage().getSafetyFactor();
     }
 
+    enum FormattedValueType {
```
SamBarker, line 693: Name feels clunky I'm wondering about `VolumeState` or even `VolumeFormatState`

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -666,38 +665,55 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current) {
...
+        FORMATTED, UNFORMATTED
+    }
+
+    /**
+     * Given either a formatted or unformatted size, calculate the formatting overhead using a heuristic.
+     */
+    protected long calculateFormattingOverhead(ManagedKafka managedKafka, long size, FormattedValueType formattedValueType) {
```
SamBarker, line 700: Given the `formattedValueType` is determined at the call site, I wonder if it's simpler to have two methods and easier to read. 

`calculateFormattedSize` & `calculateUnformattedSize`? 

That gets rid of the if branch and avoids the need for the enum.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -666,38 +665,55 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current) {
...
+        }
+    }
+
     /**
      * Get the effective volume size considering extra padding and the existing size
      */
-    private Quantity getAdjustedMaxDataRetentionSize(ManagedKafka managedKafka, Kafka current) {
-        long bytes = getPerBrokerBytes(managedKafka, current, managedKafka.getSpec().getCapacity().getMaxDataRetentionSize(), () -> this.configs.getConfig(managedKafka).getKafka().getVolumeSize());
+    private Quantity getAdjustedMaxDataRetentionSize(Kafka current, long storagePerBroker, long storagePadding) {
+        long bytes = storagePerBroker;
```
SamBarker, line 713:
```suggestion
        long bytes = storagePerBroker + storagePadding;
```

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -666,38 +665,55 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current) {
...
 
         // pad to give a margin before soft/hard limits kick in
-        bytes += getStoragePadding(managedKafka, current);
+        bytes += storagePadding;
```
SamBarker, line 716:
```suggestion

```

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
```diff
@@ -130,13 +131,13 @@ void testManagedKafkaToKafkaWithSizeChanges() throws IOException {
 
         Kafka reduced = kafkaCluster.kafkaFrom(exampleManagedKafka("40Gi"), kafka);
 
-        // should not change to a smaller size
-        diffToExpected(reduced, "/expected/strimzi.yml");
+        // storage should not change to a smaller size, but the lower limits should be enforced
```
SamBarker, line 134: OOH that's a good spot!

SamBarker (approved):
I've no issues with the substance of the change, just a few comments on the "style" or the change. 

No objections to merging as is.

shawkins:
> Steve now sure what you suggest should be documented. I wasn't expected any of this bugfix/implementation to result in the user docs.

@bibryam I'm suggesting that max data retention size is documented as volume size.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: Agreed

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: I wondered about this myself.  I reasoned that "something" might prevent the system reaching its desired number broker, so using actual for the calculation seemed to be better.  I don't feel strongly about this.  

I propose to leave it as it is.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: done

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: the enum is removed

shawkins (approved):
LGTM

PR #707, opened 2022-04-08
MGDSTRM-7790 adding a delay to modifying the ingress deployment

shawkins:
This adds:

* a delay - since we observe events firing rapidly together from the deployment, before our edit has taken effect (there will be upstream changes that help introduce an application level cache that can prevent this in most circumstances in the future)
* a check if the edit is needed - I have not reproduced the situation or observed directly the cause, but it's possible that even what we think is an empty edit/patch could still be updating the resource, so we should check before hand if it really needs updated.


rareddy (approved):
lgtm

Do you want to edit the JIRA to add in the additional work being done to reset the resources?

shawkins:
> Do you want to edit the JIRA to add in the additional work being done to reset the resources?

I'll add the description from this pr as a comment on the JIRA.

MikeEdgar (approved):
lgtm

PR #708, opened 2022-04-14
upgrading to quarkus operator sdk 3.0.7

shawkins:
This appears to be the next versions that will be productized.  Do not merge - until this has been tested in container and we have confirmation on the versions.


shawkins:
The next commits

* remove the finalizer from managedkafka/managedkafkaagent.  It includes changes to remove the finalizer from existing instances, which we can later remove.
* remove startup annotations from beans that do not have real logic in their postconstruct methods.

Tested the changes in an actual instance without an issue.  Will try a few more restarts just to be sure there's nothing weird still going on with the configuration.

pom.xml
```diff
@@ -56,16 +56,16 @@
 
         <maven.compiler.parameters>true</maven.compiler.parameters>
 
-        <quarkus.platform.version>2.2.2.Final</quarkus.platform.version>
+        <quarkus.platform.version>2.6.2.Final</quarkus.platform.version>
```
MikeEdgar, line 59: I think the upcoming product release is 2.7.5. Does that change anything with the other dependency version updates?

shawkins: > I think the upcoming product release is 2.7.5. Does that change anything with the other dependency version updates?

Yes, 2.7.5 is the intended final state.  We're waiting to hear back on support of java operator sdk 3.0.7 and fabric8 5.12.2

shawkins: See the next commit - switching to the 3.0.x sdk is a larger change that I haven't fully vetted.  There are some additional follow on items this does offer - such as access to the primary resource cache.

rareddy:
there seem to be some concerns around moving to a GA version and not being able to support CVE as the release cycle for Quarkus is 6 months long. more discussion is needed here..

rareddy (approved):
lgtm

MikeEdgar (approved):
Looks good - ran it via OLM bundle in kas-installer and no issues as of the update to application.properties. Thanks @shawkins

shawkins:
It's probably not clear from the pr, but the only thing that we are loosing is that we are no longer being provided with the list of events that trigger the controller reconciliation.  If need be we can open an upstream issue on what log level is needed in their stuff to see that, or to add it back to the reconciliation call.

MikeEdgar:
> It's probably not clear from the pr, but the only thing that we are loosing is that we are no longer being provided with the list of events that trigger the controller reconciliation.

I noticed that it was removed from the logging, but didn't think to look for changes to the `Context`. Would additional logging (at different levels perhaps) in `ResourceEventSource` fill the gap?

shawkins:
> Would additional logging (at different levels perhaps) in ResourceEventSource fill the gap?

It helps, but it's more disjointed.  There is a log down in the sdk EventProcessor.handleEvent that should be one-for-one with our ResourceEventSource, but nothing that I can find that aggregates them per reconciliation.  Based upon the logic in EventMarker they no longer try to accumulate the events per reconciliation - just track if reconciliation is needed.

shawkins:
There will eventually be a separate followup to use the sdk's framework for dependent resources.

PR #712, opened 2022-04-18
Moving out from the managedKafka CR to secrets

biswassri:
[MGDSTRM-5205](https://issues.redhat.com/browse/MGDSTRM-5205)

bootstrap certificates
oauth - clientid, clientSecret
Canary service account

[Replacing [https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/699] ]


common/src/main/java/org/bf2/common/OperandUtils.java
```diff
@@ -46,6 +47,12 @@ public static Map<String, String> getDefaultLabels() {
         return result;
     }
 
+    public static Map<String, String> getMasterSecretLabel() {
+        LinkedHashMap<String, String> result = new LinkedHashMap<>(1);
+        result.put(MANAGED_BY_LABEL, MASTER_SECRET_NAME);
+        return result;
+    }
```
MikeEdgar, line 54: I see that this method is used for filtering the secrets used by the informer, but the labels are not being set on the secrets. Rather than `managed-by`, I wonder if setting something like `app.kubernetes.io/component=master-secret` would make more sense for this secret.

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

sync/src/main/java/org/bf2/sync/SecretManager.java
```diff
@@ -0,0 +1,221 @@
...
+
+    Secret cachedOrRemoteSecret(ManagedKafka managedKafka, String name) {
+        Secret secret = cachedSecret(managedKafka, name);
+
+        if (secret == null) {
+            secret = secretResource(managedKafka, name).get();
+        }
+
+        return secret;
+    }
```
MikeEdgar, line 210: Unused method

operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
```diff
@@ -98,8 +104,10 @@ public void createOrUpdate(ManagedKafka managedKafka) {
         Secret currentSsoTlsSecret = cachedSecret(managedKafka, ssoTlsSecretName(managedKafka));
 
         if (isKafkaAuthenticationEnabled(managedKafka)) {
+            Secret ssoClientSecret = ssoClientSecretFrom(managedKafka, currentSsoClientSecret, currentMasterSecret);
+            createOrUpdate(ssoClientSecret);
```
MikeEdgar, line 108: I think something with this part needs to change. Perhaps have `ssoClientSecretFrom` also check that `managedKafka.getSpec().getOauth().getClientSecret()` is not null before using it. Otherwise return `null` and call `createOrUpdate` only when a nonnull value was returned. 

The check in `AbstractKafkaCluster#buildSsoClientGenericSecretSource` should also check that the clientSecretRef is null before returning null.

MikeEdgar (changes requested):

common/src/main/java/org/bf2/common/OperandUtils.java
rareddy: +1

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -152,15 +156,18 @@ boolean deleteAllowed(ManagedKafka local) {
...
             log.warnf("Ignoring ManagedKafka %s that wants to come back to life", Cache.metaNamespaceKeyFunc(existing));
             return false;
         }
-
-        return !remoteSpec.equals(existing.getSpec()) || !Objects.equals(existing.getMetadata(), remoteManagedKafka.getMetadata());
+        if (secretManager.isMasterSecretChanged(remote, existing)) {
+            return true;
+        }
+        ManagedKafka remoteCopy = secretManager.removeSecretsFromManagedKafka(remote);
+        return !remoteCopy.getSpec().equals(existing.getSpec()) || !Objects.equals(existing.getMetadata(), remote.getMetadata());
```
rareddy, line 170: `remote.getMetadata()` maybe changed to `remoteCopy.getMetadata()` for consistency

sync/src/main/java/org/bf2/sync/informer/InformerManager.java
```diff
@@ -46,6 +51,19 @@ protected void onStart() {
         managedAgentInformer = resourceInformerFactory.create(ManagedKafkaAgent.class, client.resources(ManagedKafkaAgent.class).inAnyNamespace(),
                 CustomResourceEventHandler.of(controlPlane::updateAgentStatus));
 
+        secretInformer = resourceInformerFactory.create(Secret.class, client.secrets().inAnyNamespace().withLabels(OperandUtils.getMasterSecretLabel()),
+                new ResourceEventHandler<Secret>() {
```
shawkins, line 55: This can just be null.

common/src/main/java/org/bf2/common/OperandUtils.java
shawkins: component would be fine, or managed/created-by sync.

sync/src/main/java/org/bf2/sync/SecretManager.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        // Sort by key to ensure iteration order for any input map
+        data.entrySet()
+            .stream()
+            .sorted(Comparator.comparing(Map.Entry::getKey))
+            .forEach(entry ->
+                secretsDigest.update(entry.getValue().getBytes(StandardCharsets.UTF_8)));
+
+        return String.format("%040x", new BigInteger(1, secretsDigest.digest()));
```
shawkins, line 199: would it be possible to extract some common secret logic between this and SecuritySecretManager?

shawkins (commented):
I think this is looking good.  Just a few comments.

common/src/main/java/org/bf2/common/OperandUtils.java
biswassri: Changed it to `app.kubernetes.io/component=master-secret`

sync/src/main/java/org/bf2/sync/informer/InformerManager.java
biswassri: Do you mean the inside the call for the new `ResourceEventHandler` or the type of `ResourceEventHandler` ?

MikeEdgar (approved):
I think this is good after conflict resolution/squashing. Thanks @biswassri

shawkins (approved):
LGTM, just a minor comment.

sync/src/main/kubernetes/kubernetes.yml
```diff
@@ -33,6 +33,18 @@ rules:
       - delete
       - update
       - patch
+  - apiGroups:
```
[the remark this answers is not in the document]
rareddy: we have the same scenario with operator permissions too, I remember raising that as a concern but it seems like there was no alternative as the access was required across the namespaces for the operator. I would have preferred if Kube based it on some kind of selector TBH

I doubt prodsec went at that level of introspection

api/src/main/java/org/bf2/operator/resources/v1alpha1/SecretKeySelector.java
```diff
@@ -0,0 +1,88 @@
...
+        "name",
+        "optional"
+})
+@ToString
+@EqualsAndHashCode
+@Buildable(editableEnabled = false, validationEnabled = false, lazyCollectionInitEnabled = false, builderPackage = "io.fabric8.kubernetes.api.builder")
+public class SecretKeySelector implements KubernetesResource
+{
+
+    @JsonProperty("key")
```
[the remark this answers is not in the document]
biswassri: Right! Makes sense.

rareddy (approved):
lgtm

PR #713, opened 2022-04-19
Release 0.22.0

MikeEdgar:
This PR also goes back to having the `main` POM versions be set to the next SNAPSHOT release (see #230), as set by `next-version` for each release.

@rareddy , @shawkins - we can also leave `next-version` set to `999-SNAPSHOT`, let me know your thoughts.


rareddy (approved):
lgtm

PR #716, opened 2022-04-22
copying only the labels and annotations

shawkins:
The operator sdk will not process events if the finalizer is removed (which seems like an sdk bug, which will no longer apply to us after https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/708/commits/e7f77ab96da923d068ef4e12edeee585dd1c6403)

So changing the metadata copying to be just labels and annotations.

Also this could have been avoided with fabric8 apply support - however even basic server side apply support is not coming until 6.0.


MikeEdgar (approved):
LGTM, verified the before/after locally.

rareddy (approved):
lgtm

PR #718, opened 2022-04-27
fix #660: removing set -e as that causes the shell to later terminate

### LINKED ISSUE #660: osd-provision script terminates shell on error
If you run perf/scripts/osd-provision.sh and something goes wrong, your whole terminal session will go away.  The workaround is to call the script as a background process.

shawkins:
See #660

### DIFF
--- a/perf/scripts/update_env.sh
+++ b/perf/scripts/update_env.sh
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 # Usage: . update_env.sh <AWS credentials CSV> <region>
 
-set -e
-
 SED=sed
 GREP=grep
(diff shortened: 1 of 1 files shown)


MikeEdgar:
Is this script missing some conditional logic that was causing it to have the error?

shawkins:
> Is this script missing some conditional logic that was causing it to have the error?

In this case the effect of the set -e bleed over into the calling shell and then if you tried to do something benign that returned a non-zero exit, it caused the shell to terminate.

MikeEdgar (approved):

rareddy (approved):
lgtm

PR #719, opened 2022-04-27
Configure new version of the quota plugin

k-wall:
The new quota plugin requires additional configuration options. This PR passes those options.   The new configuration options will be ignored by the older version of the plugin, so will be safe for older bundle versions.

This PR also let's you override the kafka/zookeeper images via either the configmap or fleetshard environment variables.  Just set a env var on the deployment like this and it will override the image provided by the rhosak bundle.

```
 oc set env deployment -n  redhat-kas-fleetshard-operator kas-fleetshard-operator IMAGE_KAFKA=quay.io/k_wall/kafka-30@sha256:098303da416359ae1ff6509af047ae9af004015e91f57f59d0b37ddbb8da8f80 
```


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -613,8 +616,13 @@ private void addQuotaConfig(ManagedKafka managedKafka, Kafka current, Map<String
...
         config.put("client.quota.callback.static.storage.hard", String.valueOf(storageLimit));
 
+
         // Check storage every storageCheckInterval seconds
         KafkaInstanceConfiguration instanceConfig = this.configs.getConfig(managedKafka);
+        if (instanceConfig.getKafka().isEnableQuota()) {
+            config.put("client.quota.callback.usageMetrics.topic", String.valueOf(instanceConfig.getKafka().getQuotaCallbackUsageMetricsTopic()));
+            config.put("client.quota.callback.quotaPolicy.check-interval", String.valueOf(instanceConfig.getKafka().getQuotaCallbackQuotaPolicyCheckInterval()));
+        }
         config.put("client.quota.callback.static.storage.check-interval", String.valueOf(instanceConfig.getStorage().getCheckInterval()));
```
SamBarker, line 626: Is there a reason not to include this property in the above if condition?

k-wall: fixed

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -220,6 +221,7 @@ public Kafka kafkaFrom(ManagedKafka managedKafka, Kafka current) {
                         .withAuthorization(buildKafkaAuthorization(managedKafka))
                         .withImage(kafkaImage.orElse(null))
                         .withExternalLogging(buildKafkaExternalLogging(managedKafka))
+                        .withImage(this.overrideManager.getKafkaImage(strimzi).orElse(null))
```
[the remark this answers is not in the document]
k-wall: That was an oversight, it is fixed.

grdryn (approved):

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -617,7 +618,13 @@ private void addQuotaConfig(ManagedKafka managedKafka, Kafka current, Map<String
 
         // Check storage every storageCheckInterval seconds
         KafkaInstanceConfiguration instanceConfig = this.configs.getConfig(managedKafka);
-        config.put("client.quota.callback.static.storage.check-interval", String.valueOf(instanceConfig.getStorage().getCheckInterval()));
+        if (instanceConfig.getKafka().isEnableQuota()) {
+            KafkaInstanceConfiguration.Kafka kafka = instanceConfig.getKafka();
+            config.put("client.quota.callback.usageMetrics.topic", String.valueOf(kafka.getQuotaCallbackUsageMetricsTopic()));
+            config.put("client.quota.callback.quotaPolicy.check-interval", String.valueOf(kafka.getQuotaCallbackQuotaPolicyCheckInterval()));
+            config.put("client.quota.callback.kafka.clientIdPrefix", String.valueOf(kafka.getQuotaCallbackQuotaKafkaClientIdPrefix()));
+            config.put("client.quota.callback.static.storage.check-interval", String.valueOf(instanceConfig.getStorage().getCheckInterval()));
```
MikeEdgar, line 626: Just to mention it, the `OperandOverride` class supports `additionalProperties` that can be used to configure properties specific to a Strimzi bundle deployment version. 

That can be leveraged to only apply new/changed properties alongside the image updates they align with. I think ultimately we'll want to use that (or something like it) to extend support for blast-radius control to include configuration as well.

MikeEdgar (approved):

k-wall:
@SamBarker can you re-review, once you are happy I believe this is good to merge.

SamBarker (approved):
LGTM

PR #720, opened 2022-04-28
Add config to enable partition limit enforcement

grdryn:
JIRA: https://issues.redhat.com/browse/MGDSTRM-8361

### DIFF
--- a/operator/src/main/resources/instances/managedkafka.properties
+++ b/operator/src/main/resources/instances/managedkafka.properties
@@ -24,3 +24,6 @@ storage.safety-factor=2
 storage.min-margin=10Gi
 # Based on experimentation with ext4 volumes (the kubernetes default which is used by the service).
 storage.formatting-overhead=0.05
+
+# Override for partition limit enforcement. This override should be removed once we can enable this by default
+kafka.partition-limit-enforced=false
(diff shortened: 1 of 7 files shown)


operator/src/main/resources/instances/managedkafka.properties
```diff
@@ -24,3 +24,6 @@ storage.safety-factor=2
 storage.min-margin=10Gi
 # Based on experimentation with ext4 volumes (the kubernetes default which is used by the service).
 storage.formatting-overhead=0.05
+
+# Feature flag to toggle enabling of partition enforcement
+kafka.partition-limit-enforced=false
```
grdryn, line 29: @k-wall I might be over-thinking this: essentially, this PR could be merged _now_, then when we want to turn on enforcing, we just remove this override. WDYT?

shawkins: @grdryn is this property something that you see changing per profile type, per strimzi version, or both?

grdryn: @shawkins I _think_ it would be changed for everything at once (this is why I say I might be over-thinking it by having a property here to configure it...it might be simpler to just have a hard-coded value that's changed on an FSO release)?

@k-wall as you suggested it, do you have any thoughts or preferences around which approach you'd prefer?

k-wall: For production use-cases, I don't think we'd ever want to turn this off.  I suppose it might be useful to keep the flag to support developer use-cases.  Long term, I would make the default true and omit from the properties files.

The reason we need to turn it off the now it that we have users already over the partitions limit (the KafkaClusterPartitionsLimitExceeded is firing.  Once the partition limit is revised upwards, the customer will be under the limit, we should promptly enable the feature to prevent future breaches.

grdryn: Thanks @k-wall, I've updated the comment now.

@shawkins let me know if this override should be placed somewhere else than this file, or if you have any other questions or feedback. Thanks!

k-wall (approved):

SamBarker (approved):
LGTM

grdryn:
@shawkins @rareddy I don't have the ability to merge this. If you're happy with it, could you please merge it?

PR #721, opened 2022-05-02
Add ability to change display name into osd-provision.sh

kornys:


MikeEdgar (approved):

perf/scripts/osd-provision.sh
```diff
@@ -26,6 +26,7 @@ function usage() {
         --install-addon ADDON_ID                    install selected addon id
         --remove-addon ADDON_ID                     uninstall selected addon id
         --set-ingress-controller                    setup ingresscontroller
+        --set-display-name DISPLAY_NAME             change display name of the cluster (it does not change name or od of cluster)
```
shawkins, line 29:
```suggestion
        --set-display-name DISPLAY_NAME             change display name of the cluster (it does not change name or id of cluster)
```

k-wall (approved):

grdryn (approved):
Is this expected to work as both an option when creating new clusters, and as a modifier for existing clusters? I just attempted to use it to modify the display name of my existing cluster, and while it didn't error, it didn't seem to have any effect. Maybe I just need to wait longer for the change to appear?

shawkins (approved):
LGTM

PR #724, opened 2022-05-05
MGDSTRM-8370 adding affinity when there are multiple profiles

shawkins:
This adds Profiles to the ManagedKafkaAgent.  For now they are empty, but it's expected that with capacity reporting that they will at least contain the max nodes.  Here we're using the presence of both developer and standard to trigger adding nodeaffinity (effectively a nodeselector) to all the pods.

There may eventually be a map of profiles, rather than just well known names, but for now we are effectively building our solution around having exactly these named types.

This may also need to be generalized further to adding a toleration.


shawkins:
The second commit adds support for tolerations.  These will always be added as long as there is a profile set.  Now that we have moved away from eval instances at some point we could go back and make the profile required.

common/src/main/java/org/bf2/common/OperandUtils.java
```diff
@@ -79,15 +88,74 @@ public static <T extends HasMetadata> T createOrUpdate(MixedOperation<T, ?, ?> r
...
+                .withKey(ManagedKafka.PROFILE_TYPE)
+                .withValue(type)
+                .withEffect("NoExecute")
+                .build());
+    }
+
+    public static NodeAffinity nodeAffinity(ManagedKafkaAgent agent, ManagedKafka managedKafka) {
+        if (agent == null || agent.getSpec().getCapacity().size() <= 1) {
+            return null; // not expected to use a node label
+        }
```
rareddy, line 112: Is the default machine pool used over the named machine pool in a single profile cluster? I thought if we use auto-scaler then we use named pools.

was also thinking would it be better to extract this code as an interface method.

shawkins: This check along with the code that deals with the profile being absent are mostly for legacy / backwards compatibility.  As we move forward these cases won't actually occur in staging / production - it will be required for the agent to be configured with the appropriate profiles and for those to be present on the managedkafka (should already be the case as evals are now gone).

operator/src/test/java/org/bf2/operator/operands/AdminServerTest.java
```diff
@@ -82,12 +77,17 @@ static ManagedKafka buildBasicManagedKafka(String name, String strimziVersion, T
...
+        "test-mk-tls-q, 2, true,  true, false, /expected/adminserver-tls.yml",
+        "test-mk-q,     2, true,  false, true, /expected/adminserver-affinity.yml"
     })
-    void createAdminServerDeployment(String name, String versionString, boolean quarkusBased, boolean tls, String expectedResource) throws Exception {
+    void createAdminServerDeployment(String name, String versionString, boolean quarkusBased, boolean tls, boolean useNodeAffinity, String expectedResource) throws Exception {
         ManagedKafka mk = buildBasicManagedKafka(name, versionString, tls ? new TlsKeyPair() : null);
 
+        if (useNodeAffinity) {
+            CanaryTest.useNodeAffinity(mk);
+        }
```
MikeEdgar, line 89: This seems a little confusing. Might be better to put `useNodeAffinity` in some test util class or similar.

shawkins: When everything is in the same package I'm not too concerned about having separate utility classes.  If you prefer one, it can of course be added.

MikeEdgar: Thanks for adding that, looks good.

MikeEdgar (approved):

common/src/main/java/org/bf2/common/OperandUtils.java
rareddy: Thanks

rareddy (approved):
lgtm

PR #725, opened 2022-05-09
MGDSTRM-8409 cut down on modifications to the ingresscontroller

shawkins:
Prior to issuing a patch we'll use the diff logic to determine if the patch is even needed.  Simply having fields added in existing state is just a side effect of the ingress controller operator - with the worst case is that the application of our patch then causes another update, which can repeat for quite a while.

So we need to see something removed or modified to submit the patch.  

I only see two updates being clustered by the operator in my local testing, so I didn't bother adding a time delay like with the ingress replicas.

This strategy could be employed for any dependent resource that we are attempting to patch that is modified by their respective controller / operator.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -1,5 +1,7 @@
 package org.bf2.operator.managers;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.fabric8.kubernetes.api.builder.TypedVisitor;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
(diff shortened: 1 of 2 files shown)


operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -395,22 +399,47 @@ void reconcileIngressControllers() {
...
+    boolean shouldPatchIngressController(IngressController expected, IngressController existing) {
+        ObjectMapper objectMapper = Serialization.yamlMapper();
+        JsonNode expectedJson = objectMapper.convertValue(expected, JsonNode.class);
+        JsonNode actualJson = objectMapper.convertValue(existing, JsonNode.class);
+        JsonNode patch = JsonDiff.asJson(expectedJson, actualJson);
+        if (patch.isArray()) {
+            int size = patch.size();
+            for (int i = 0; i < size; i++) {
+                JsonNode op = patch.get(i).get("op");
+                if (op != null && !"add".equals(op.asText())) {
```
rareddy, line 434: so basically you are expecting no one to change the CR for the ingress controller right? I mean additions, but expect to be modified as removed segments of it

shawkins: We're guarding against any changes here - someone can modify it, but we want to make sure that the spec and label changes we want are there.  If there are additions that's okay - primarily that accounts for the httpEmptyRequestsPolicy field in the spec, which is being dropped from the expected version by https://github.com/fabric8io/kubernetes-client/issues/3570 

If there are modifications or removals though, that means that the current version has changes wrt to our expected version that we want to overwrite.

grdryn (approved):

PR #734, opened 2022-05-31
Adding test for annotation

biswassri:
Small fix to the secret manager to update secret-digest and tests for the annotation of the `master-secret`

PR #735, opened 2022-05-31
a follow up to #734 - allowing the fleetshard to set annotations

shawkins:
Our logic to detect changes and retain data plane annotations needed refinement.

We can't compare the entire metadata as that will look not only at all annotations, but at the kubernetes internal metadata fields - so the previous version was effectively always returning true.

Then to properly account for annotations managed by the data plane vs those maintained by the control plane we need to maintain a set of what we might add / change and then account for that in the change detection and in the update to the local managedkafka.


sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -231,7 +245,12 @@ void reconcile(String remoteId, String localMetaNamespaceKey) {
                     ObjectMeta meta = remoteCopy.getMetadata();
                     client.edit(local.getMetadata().getNamespace(), local.getMetadata().getName(), mk -> {
                             mk.getMetadata().setLabels(meta.getLabels());
+                            Map<String, String> existingAnnotations = mk.getMetadata().getAnnotations();
                             mk.getMetadata().setAnnotations(meta.getAnnotations());
+                            if (existingAnnotations != null) {
+                                existingAnnotations.keySet().retainAll(DATA_PLANE_ANNOTATIONS);
+                                mk.getMetadata().getAnnotations().putAll(existingAnnotations);
+                            }
```
MikeEdgar, line 253: Is there any chance that `existingAnnotations` is an unmodifiable map?

MikeEdgar (approved):
lgtm

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -167,7 +176,12 @@ public boolean changed(ManagedKafka remote, ManagedKafka existing) {
             return true;
         }
         ManagedKafka remoteCopy = secretManager.removeSecretsFromManagedKafka(remote);
-        return !remoteCopy.getSpec().equals(existing.getSpec()) || !Objects.equals(existing.getMetadata(), remoteCopy.getMetadata());
+        // will always be non-null due to the id/placement requirements
+        Map<String, String> annotations = new HashMap<>(existing.getMetadata().getAnnotations());
+        annotations.keySet().removeAll(DATA_PLANE_ANNOTATIONS);
```
biswassri, line 181: Are we removing all the data-plane annotations here to detect if any control plane annotations have changed?

biswassri (approved):

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
shawkins: Yes.  They aren't doing that yet we're just future-proofing.  The placement id annotation can change, but the exiting managedkafka should be deleted before they make that change.

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -231,7 +245,12 @@ void reconcile(String remoteId, String localMetaNamespaceKey) {
                     ObjectMeta meta = remoteCopy.getMetadata();
                     client.edit(local.getMetadata().getNamespace(), local.getMetadata().getName(), mk -> {
                             mk.getMetadata().setLabels(meta.getLabels());
+                            Map<String, String> existingAnnotations = mk.getMetadata().getAnnotations();
                             mk.getMetadata().setAnnotations(meta.getAnnotations());
+                            if (existingAnnotations != null) {
+                                existingAnnotations.keySet().retainAll(DATA_PLANE_ANNOTATIONS);
+                                mk.getMetadata().getAnnotations().putAll(existingAnnotations);
```
rareddy, line 252: I am failing to read this (never used retainAll method before), so we are removing all the annotations from existing MK except for DATA_PLANE_ANNOTATIONS, and updating the resource. What happens to any other annotations that are set by CP or Kube?

rareddy (approved):
lgtm

PR #736, opened 2022-05-31
Use new prefix for broker plugins properties

grdryn:
In the broker plugins project, we're updating the config properties so that they don't have the strimzi prefix.

This change updates the config set by this operator so that both the old and new properties are set. Once the version of the operator using the new properties is rolled out, we can then remove the old properties from being set here again in another change.

Given that this is intended to be a temporary duplication, I tried to keep it as simple as possible, so that it will be simple to remove when we get to that stage also.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
@@ -388,6 +388,8 @@ public class KafkaInstanceConfiguration {
         protected String finalVersion = "0.0.0"; // Default, disables `acl-legacy` configurations
         @JsonProperty("authorizer-class")
         protected String authorizerClass = null;
+        @JsonProperty("broker-plugins-config-prefix")
+        protected String brokerPluginsConfigPrefix = null;
         @JsonProperty("config-prefix")
         protected String configPrefix = null;
         @JsonProperty("global")
@@ -457,10 +459,20 @@ public class KafkaInstanceConfiguration {
             this.authorizerClass = authorizerClass;
         }
 
+        public String getBrokerPluginsConfigPrefix() {
+            return brokerPluginsConfigPrefix;
+        }
+
+        public void setBrokerPluginsConfigPrefix(String brokerPluginsConfigPrefix) {
+            this.brokerPluginsConfigPrefix = brokerPluginsConfigPrefix;
+        }
+
+        @Deprecated
         public String getConfigPrefix() {
             return configPrefix;
         }
 
+        @Deprecated
         public void setConfigPrefix(String configPrefix) {
             this.configPrefix = configPrefix;
         }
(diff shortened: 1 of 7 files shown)


operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
```diff
@@ -390,6 +390,8 @@ public static class AccessControl {
         protected String authorizerClass = null;
         @JsonProperty("config-prefix")
         protected String configPrefix = null;
+        @JsonProperty("old-config-prefix")
+        protected String oldConfigPrefix = null;
```
SamBarker, line 394: Not a fan of using `oldConfigPrefix` if there is any chance of this change hanging around, but I'm struggling to find a better name. 

One alternative that comes to mind is given this is part of  `KafkaInstanceConfiguration` (technically its in the `AccessControl` class so maybe the scope is already explicit enough?) maybe the new property should be called  `brokerPluginsConfigPrefix` to distinguish it from config for other plug-ins?

SamBarker (approved):
LGTM

I don't see a easier way of managing the transition.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
grdryn: Sounds good, I'll update to that and rebase (it's currently conflicting)

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
```diff
@@ -388,6 +388,8 @@ public static class AccessControl {
         protected String finalVersion = "0.0.0"; // Default, disables `acl-legacy` configurations
         @JsonProperty("authorizer-class")
         protected String authorizerClass = null;
+        @JsonProperty("broker-plugins-config-prefix")
+        protected String brokerPluginsConfigPrefix = null;
         @JsonProperty("config-prefix")
         protected String configPrefix = null;
```
MikeEdgar, line 394: Is this deprecated now?

MikeEdgar (approved):
One (non blocking) question, otherwise looks good. I think eventually we'll need to put these configs into the Strimzi bundle's configmap so they only apply to the specific bundle versions.

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
grdryn: Ah yes, I guess so. I'll add the annotation to it now, thanks!

grdryn:
> I think eventually we'll need to put these configs into the Strimzi bundle's configmap so they only apply to the specific bundle versions.

@MikeEdgar I'm not sure about this part. The intention of the duplication at the moment is to allow for this to be released independent of the version of the plugin in the broker. Then at some point in the future when there are no instances of the older plugin version remaining, the old config can be removed.

I'm probably misunderstanding what you're saying...could you point me in the direction of the configmap that you're referring to? That might make it clearer to me :slightly_smiling_face:

PR #737, opened 2022-06-02
Initial logic for actual capacity reporting

shawkins:


shawkins:
I'll add a few more tests, but this does seem to be functionally complete.  The idea is that the controllers will need to maintain a configmap that tracks the resource utilization - for each ManagedKafka and the total overall.  Modifications to this configmap are done in a locked manner to ensure we are reasoning over it in a serialized fashion.  For now we are making the assumption that the resource attributes of a managedkafka are fixed - if/when they are fungible the logic will need to change.  I'm also using the quota consumed label for the number of units, rather than the calculation for the number of brokers that we have based upon the partitions.  We should consider going back and simplifying the other logic to use the quota consumer label as well - after all the discussions around capacity planning and related topics it was clear that we're pretty tied in the near/intermediate term to the idea that the quota consumed will be 1 for 1 with the physical units.

As implemented the corner cases are:
- someone directly modifies the resources configmap - we could consider validating / rebuilding the configmap on the reconciliation loop for the ManagedKafkaAgent
- a ManagedKafka is removed when the operator is not running - that's not expected to happen, but could be resolved in a similar way as the first point.

shawkins:
Added logic to check for orphaned entries.  Nothing yet for the case where an admin directly modifies the resource configmap.

rareddy:
@shawkins when looking through the code, I felt uneasy using the config map to keep the state and trying to manage it, rather than using it as input. I was thinking, why not decorate the MKA's status with the same information? I know we send the MKA Status back to CP, but we could strip if we think it is going to create a large blob (IMO it should be ok, as this list might ever glow in 100s)

I am still on the fence about using size based on quota label.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,335 @@
...
+
+    static Resources createResources(ManagedKafka mk, String currentProfile) {
+        Resources resources = new Resources();
+        resources.profile = currentProfile;
+        resources.units = Integer.parseInt(
+                OperandUtils.getOrDefault(mk.getMetadata().getLabels(), ManagedKafka.PROFILE_QUOTA_CONSUMED, "1"));
+        return resources;
+    }
+
+    public void releaseResources(ManagedKafka managedKafka) {
```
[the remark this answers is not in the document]
shawkins: Yes it can be.  The intent of the synchronization is to cut down on the need for retries.

shawkins:
> I felt uneasy using the config map to keep the state

That's precisely what configmaps are for - state.

> I was thinking, why not decorate the MKA's status with the same information?

Status by kubernetes design is not supposed to be state - it can be lost / recreated.  Even with this pr I cheating a bit in that regard - once something is marked as rejected it stays rejected.  In theory we should be tracking that through something other than the status.  

> I know we send the MKA Status back to CP, but we could strip if we think it is going to create a large blob (IMO it should be ok, as this list might ever glow in 100s)

There are two patterns for holding onto additional state - on a resource via an annotation (such as how kubectl manages the last state for apply) or via a configmap / secret.  I'd rather use a separate configmap than to try to manage this as annotations on the managedkafkaagent.  I also don't want to use both annotations on the managedkafkas (for tracking their approved resource configuration) and the aggregate information on something else - it's much easier to make atomic updates on just one resource.

> I am still on the fence about using size based on quota label.

We've really made things a bit of a mess in that regard...  it would make things quite a bit simpler if we adopted this - not just here but in the other operator logic.  Computing the number of brokers based upon partitions adds little value at this point.

rareddy:
Maybe I am missing the point, was there any reason to keep this state, why not calculate capacity each reconciliation cycle? as I see it the code is already walking through all the instances to find the orphans.

shawkins:
> Maybe I am missing the point, was there any reason to keep this state, why not calculate capacity each reconciliation cycle? 

There are multiple reasons:
- the "good" state for the resources is not being saved anywhere.  Rather than trying to infer what type of deployment it is from the kafka, it's much simpler to capture that directly from the managedkafka.  We have two choices either save that as a fleetshard managed annotation on the managedkafka, or save it somewhere else.  I've opted for the somewhere else, which is the same place as we're maintaining the aggregate values - this avoids any corner cases where you update one of the resources, but the update of the other fails.
- the primary resource cache of the operator sdk is hard to utilize from somewhere other than the ManagedKafkaControlller.  To keep the logic more encapsulated would likely mean starting our own ManagedKafka informer.
- even if we use our own informer you are creating an n^2 problem - for each reconciliation of a ManagedKafka you have to walk through every ManagedKafka.  For some of our internal events we do trigger the reconciliation of all managedkafkas.  This is not a problem yet (especially when the ManagedKafkas are in memory) for the number of managed kafkas we are dealing with but it doesn't scale well.

> as I see it the code is already walking through all the instances to find the orphans.

It performs that check one time at start-up.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+        Resources resources = createResources(managedKafka, profile);
+
+        int total = Integer.parseInt(resourceConfigMap.getData().getOrDefault(profile, "0"));
+
+        int newTotal = resources.units + total;
+        if (newTotal > max) {
+            log.warnf("Rejecting the deployment of %s/%s as the cluster is full",
+                    managedKafka.getMetadata().getNamespace(), managedKafka.getMetadata().getName());
+            return Optional
+                    .of(new OperandReadiness(Status.False, Reason.Rejected, "Cluster has insufficient resources"));
```
machi1990, line 228: @VanillaSpoon this is the status that you'll be interested in intepreting:
- Status: `Ready`, `False`
- Reason: `Rejected`
- Message: `Cluster has insufficient resources`

VanillaSpoon: Thanks @machi1990

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+        checkedForOrphans = true;
+        // may need to do this periodically as well - or based upon an informer
+        Map<String, String> data = new LinkedHashMap<>();
+        Map<String, Integer> totals = new LinkedHashMap<>();
+
+        try {
+            informerManager.createKafkaInformer();
+        } catch (KubernetesClientException e) {
+            // no bundle is installed, lookups for kafkas will return null
+        }
```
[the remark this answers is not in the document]
shawkins: No, this is to cover the case where the operator is installed, but strimzi is not.  It also helps with testing so you don't have to have that call in the tests.  See the strimzimanager logic that normally triggers the creation.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+            Resources resources = createResources(mk, currentProfile);
+            totals.compute(currentProfile, (k, v) -> (v == null ? 0 : v) + resources.units);
+            data.put(Cache.metaNamespaceKeyFunc(mk), Serialization.asJson(resources));
+        });
+
+        // update with the totals as well
+        totals.forEach((k, v) -> data.put(k, String.valueOf(v)));
+
+        if (data.isEmpty()) {
+            data.put("standard", "0"); // the map logic is easier if it's non-empty, util fabric8 6 we'll get back a null data
```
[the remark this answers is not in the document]
shawkins: Sure - it will also need to expose the lower name.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Responsible for ensuring the resources for a given ManagedKafka are available
+ * prior to allowing the deployment to proceed.
+ */
+@ApplicationScoped
+public class ResourceManager {
```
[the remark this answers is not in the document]
shawkins: There will be quite a few things to rename.  Let's make sure there is consensus on the name.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+        Optional<ConfigMap> resourceConfigMap = Optional.ofNullable(getCachedResourceConfigMap());
+
+        for (String key : profiles.keySet()) {
+            getProfileMaxUnits(profiles, key).ifPresent(max -> {
+                capacity.put(key,
+                        new ProfileCapacityBuilder().withMaxUnits(max)
+                                .withRemainingUnits(resourceConfigMap.map(c -> c.getData().get(key))
+                                        .map(Integer::valueOf)
+                                        .map(i -> max - i)
+                                        .orElse(null))
```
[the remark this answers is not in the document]
shawkins: My thinking here wasn't quite right - since the cached configmap is being used, it could be missing, in which case null is appropriate.  But that doesn't make sense if the map exists and just doesn't have an entry for that profile type.

It's probably better to use getOrCreateResourceConfigMap so that it's known that the configmap can't be null.  In that case the else should clearly be max.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+            log.warnf("Rejecting the deployment of %s/%s as the cluster is full",
+                    managedKafka.getMetadata().getNamespace(), managedKafka.getMetadata().getName());
+            return Optional
+                    .of(new OperandReadiness(Status.False, Reason.Rejected, "Cluster has insufficient resources"));
+        }
+
+        // update the configmap in a locked manner - this may fail, but we'll retry
+        resourceConfigMap.getData().put(profile, String.valueOf(newTotal));
+        resourceConfigMap.getData().put(Cache.metaNamespaceKeyFunc(managedKafka), Serialization.asJson(resources));
+        client.configMaps()
```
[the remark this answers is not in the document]
shawkins: This is something that fabric8 is making unnecessarily cumbersome and the code formatting makes look longer than just a single fluent call.  It actually should just be:

client.resource(resourceConfigMap).update()

At least the resource method will be available in fabric8 6.0, so some of the boiler-plate will go away.  Ideally it shouldn't really require a separate method.

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
```diff
@@ -0,0 +1,339 @@
...
+        if (newTotal > max) {
+            log.warnf("Rejecting the deployment of %s/%s as the cluster is full",
+                    managedKafka.getMetadata().getNamespace(), managedKafka.getMetadata().getName());
+            return Optional
+                    .of(new OperandReadiness(Status.False, Reason.Rejected, "Cluster has insufficient resources"));
+        }
+
+        // update the configmap in a locked manner - this may fail, but we'll retry
+        resourceConfigMap.getData().put(profile, String.valueOf(newTotal));
+        resourceConfigMap.getData().put(Cache.metaNamespaceKeyFunc(managedKafka), Serialization.asJson(resources));
```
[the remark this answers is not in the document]
shawkins: Are you talking about usage outside of ResourceManager?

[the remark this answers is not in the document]
rareddy: Not for the usage outside, but like a private class inside ResouceManager

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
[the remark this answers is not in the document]
rareddy: Is there any bean initiation ordering in Quarkus? it just looks odd that pre-conditions are not met. Even if it returns null it is not doing anything different, or does it throw a runtime exception here?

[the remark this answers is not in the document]
shawkins: > Is there any bean initiation ordering in Quarkus?

This isn't about bean ordering.  The operator is running, but no strimzi has been installed.

>  Even if it returns null it is not doing anything different, or does it throw a runtime exception here?

We are catching the exception, and just proceeding to create the configmap.  If there is an exception that means there's no kafkas, so we'll just create a default configmap.

MikeEdgar (approved):

operator/src/main/java/org/bf2/operator/managers/ResourceManager.java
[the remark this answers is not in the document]
rareddy: no big deal, but the reason I was asking for this is, the config map management and capacity building log were intertwined, I was hoping to see a CapacityManager handle the ConfigMap level but some other inner class handle all calculation, validation, reclaim kind of logic.

rareddy:
I pulled in the code locally and built local images of it, then using the "kas-installer" I deployed the instance I see the following.

Initially ControlPlane logic for supplying the "capacity" was not there, so the FSO created the configmap as below

```
data:
  standard: '0'
```
Then I updated the sync code to temporarily add "maxNodes:6" to the MKA, redeployed the operator. But the existing configmap did not change. When deleted configmap, restarted operator it showed up. ( I guess I could have waited 60secs maybe it would have updated?)

I created kafka instance I saw, following in logs

```
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://172.30.0.1/api/v1/namespaces/redhat-kas-fleetshard-operator-qe/configmaps. Message: ConfigMap "kas-fleetshard-resources" is invalid: data[kafka-cb3jq8fd70pktmpha700/foo]: Invalid value: "kafka-cb3jq8fd70pktmpha700/foo": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name', or 'KEY_NAME', or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+'). Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=data[kafka-cb3jq8fd70pktmpha700/foo], message=Invalid value: "kafka-cb3jq8fd70pktmpha700/foo": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name', or 'KEY_NAME', or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+'), reason=FieldValueInvalid, additionalProperties={})], group=null, kind=ConfigMap, name=kas-fleetshard-resources, retryAfterSeconds=null, uid=null, additional...
```

then I fixed to key having "/" to ".", and had to delete the config map, restart the operator new config map, now everything shows correctly in configmap when there is at least one Kafka instance

as
```
data:
  kafka-cb3kjofd70pktmphappg.foo: '{"profile":"standard","units":1}'
  standard: '1'
```

I deleted Delete the Kafka instance, now I see config map as 
```
data:
  standard: '0'
```
I was expecting
```
data:
  standard: '2'
``` 
As maxNodes is defined as 6. 

create kafka instance again, now the config-map updated correctly.

I only had 12 `m5.2xlarge` nodes on my cluster, but even then after the update of images I could not create successfully a kafka instance as I saw errors in pod scheduling as they said

```
0/18 nodes are available: 12 node(s) didn't match Pod's node affinity/selector, 3 node(s) had taint {node-role.kubernetes.io/infra: }, that the pod didn't tolerate, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate.```

I did not dig this into further, as I think this may be due to machine-pool which I did not create. Once I above is fixed I will do another run with machine-pool created cluster.

**Comment by shawkins** [comment]
> Then I updated the sync code to temporarily add "maxNodes:6" to the MKA, redeployed the operator.

That is the current state - as mentioned in [the other ](https://issues.redhat.com/browse/MGDSTRM-8370?focusedCommentId=20505091&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20505091) the intent is to address that in the other JIRA as we may want to put even more logic / guards against misconfiguration when the ManagedKafkaAgent spec changes.

> then I fixed to key having "/" to "."

Yes, that is an issue.  I'll need to account for this by tracking by id, rather than name.

> I was expecting

Why?  The map is tracking resources in use.

> I only had 12 m5.2xlarge nodes on my cluster, but even then after the update of images I could not create successfully a kafka instance as I saw errors in pod scheduling as they said

The expectation is that the machine pools exist with the appropriate node labels - if you are missing the node labels, then the pods will be unschedulable.

**Comment by shawkins** [comment]
> more logic / guards against misconfiguration when the ManagedKafkaAgent spec changes

To expand on that, when there is a change from manual scaling to auto-scaling or vice versa, that will cause rolling restarts of everything and possibly the need to reschedule the common machine pool pods (when going from manual to auto).  So it's on the order or a major upgrade - if for now it requires a restart of the operator that seems fine.

**Comment by shawkins** [comment]
It's probably worth mentioning here that from a managed perspective the sync component can guard against invalid transitions - as it knows both the before and desired state.  That would be similar to what we are already doing in having it guarding against invalid deletes.  However if we ever want to transition to a more generic sync component having a lot of validation / capacity logic there is probably not the best choice.  

If we were to move the logic to the sync, then we would not have to do the individual resource tracking, only the aggregates.

**Comment by rareddy** [comment]
>However if we ever want to transition to a more generic sync component having a lot of validation / capacity logic there is >probably not the best choice.

agree, let's not change that

> Why? The map is tracking resources in use.

right, my mistake, I got confused with remaining in status.

> The expectation is that the machine pools exist with the appropriate node labels - if you are missing the node labels, then the > pods will be unschedulable.

Is this going to affect existing instances? we need to be absolutely sure how we transition existing instances from default pool to the named pool. (I will spin up the cluster and do a little testing with your updated changes, in this scenario)

**Comment by shawkins** [comment]
> Is this going to affect existing instances?

Yes.  If we view this as two steps - once taints are turned on any of the common pods (including ingress replicas) will have to move to the default machine pool, if they are not there already.  Since all of those pods are higher priority than the kafka workload, it will force those pods off if need be.  Then once the managedkafkaagent is configured for multiple profiles, or has max nodes set, the operator will need to update everything to include a node selector - that should trigger a rolling restart which will also move anything that is part of the kafka workload off of the default machine pool.  

cc @machi1990

**Comment by machi1990** [comment]
> > Is this going to affect existing instances?
> 
> Yes. If we view this as two steps - once taints are turned on any of the common pods (including ingress replicas) will have to move to the default machine pool, if they are not there already. Since all of those pods are higher priority than the kafka workload, it will force those pods off if need be. Then once the managedkafkaagent is configured for multiple profiles, or has max nodes set, the operator will need to update everything to include a node selector - that should trigger a rolling restart which will also move anything that is part of the kafka workload off of the default machine pool.
> 
> cc @machi1990

ack. This matches my expectation and possibly something that will be indicated in the SOP about turning dynamic scaling on.

**Comment by rareddy** [comment]
> something that will be indicated in the SOP about turning dynamic scaling on.

@machi1990 is there SOP in works on turning on the autoscaler on and creation of machine pools for stage and prod (maybe stage for now)?

**Comment by machi1990** [comment]
> > something that will be indicated in the SOP about turning dynamic scaling on.
> 
> @machi1990 is there SOP in works on turning on the autoscaler on and creation of machine pools for stage and prod (maybe stage for now)?

No, we do not have an SOP yet. There is a task about writing an SOP but we've not gotten into it yet since it was in waiting for other work to complete first.

**Comment by rareddy** [comment]
1) I created the cluster with current versions and created a Kafka Instance.
2) Then I created a machine pool with (did not work with `NoSchedule` must be `NoExecute`)
```
ocm create machinepool --cluster rareddy-kafka --instance-type m5.2xlarge --replicas=3 --taints "bf2.org/kafkaInstanceProfileType=standard:NoExecute" --labels "bf2.org/kafkaInstanceProfileType=standard" standard
``` 
3) Updated the Operator to Steve's changes. KI did not move to machine pool based nodes (expected)
4) Modified the Sync to mimic the capacity values, still KI did not move to the machine pool based nodes
5) Deleted the canary on KI, then I saw the KI pods start moving to the machine pool based nodes. So, I guess only when an event kicks the next reconciliation cycle will it happen. Typically we manage this kind of move using blast radius so that we do not move all the instances at the same time. Here is how the event needs to be triggered and the blast radius controlled might need more discussion IMO. Or we need to make sure SRE knows about it when we deliver to the stage & prod what to expect.
6) Any next new KI correctly placed on the machine pool based nodes. Any more KI exceeding the maxNodes, I saw correctly being rejected although there is enough space on default machine pool. However, the current version probably does not understand `Rejected` status which is OK for now until I have updated Control Plans code in my `kas-installer`
7) MKA's `Status` did NOT contain the `Capacity` info in there, not sure why the code looks correct to me. This needs to be figured out. 

one observation I saw in ocm is when u delete the machinepool the nodes become part of the default pool do not get deleted.

**Comment by shawkins** [comment]
Chatted with @rareddy about this implementation vs alternatives and how our needs will evolve.  Some of the other major alternatives are:
* move the resource logic to sync, since it can act as a gate keeper and we can keep the operator more tuned to just creating the desired state
* rely on the control plane to track resource usage.  This was touched on in the epic brief, which would look more like the fleetshard advertising the resource utilization of the profiles and the max that can be placed, then the control plane would enforce the max (would be straight-forward with transactional updates to their database).  However this was not a preferred solution.
* use alternative schemes based upon prioritization - this has drawbacks as well.  
  * If you for example try to give a priority based upon creation timestamp or some sequence number that addresses that newer instances cannot take resources for very long that do not belong to them and you cannot get into a resource deadlock type situation.  However once the amount of resources become dynamic (the user can change the units after install) then this prioritization isn't meaningful - an older instance shouldn't be able to take resources from a newer.  
  * If you use a prioritization system based upon increasing the priority after successfully being ready - you still potentially have issues with resource contention among simultaneous deployments, you may have to introduce some guard to prevent advertising an instance as ready but in reality it's stealing resources (seems unlikely though) and without additional bookkeeping you still have a problem accounting for a dynamic number of units.
  * escalation of reservations, which was touched on in the epic brief - that seems to introduce even more complexity

So other than demonstrating what this logic would look like moving it to the sync, this is probably the best we can do.

**Comment by rareddy** [comment]
> Is it possible that your MKA spec is not configured for profiles, or that the operator wasn't restarted?
it is definitely configured for profiles as I saw the new instance being placed on the machinepool created as mentioned. saw the configmap so it was running latest code. I will check again.

**Comment by rareddy** [comment]
> implementation vs alternatives and how our needs will evolve
@shawkins can you make the comments also on JIRA for higher visibility perhaps, to get better participation if anyone is watching?

**Review by k-wall — APPROVED** [review]
looks reasonable to me.

**Review by rareddy — APPROVED** [review]
lgtm

**Comment by rareddy** [comment]
@shawkins one another issue I thought about is when the profile stuff is turned on, by default the machine pool Nodes will probably not be scaled enough to handle the number of instances that currently may be in existence. by reconciling we may be pushing them to wait for Node provision time along with a rolling update. Maybe we should consider adding min == max nodes first and then adjust the min value later? @machi1990

**Comment by shawkins** [comment]
> Maybe we should consider adding min == max nodes first and then adjust the min value later?

Yes, we talked about a similar approach in staging in general for turning on the auto-scaling as you can alleviate some of the slo concern by gradually dropping the min.

**Comment by machi1990** [comment]
> @shawkins one another issue I thought about is when the profile stuff is turned on, by default the machine pool Nodes will probably not be scaled enough to handle the number of instances that currently may be in existence. 

I was thinking that with or without autoscaling, the machine pool node count should be just enough to handle the current workload.
In a scenario where we do not have autoscaling turned on, the manually configured cluster is configured in a way that the node count will handle the number of kafkas limit defined in a configuration: see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/docs/data-plane-osd-cluster-options.md#using-an-existing-osd-cluster-with-manual-scaling-enabled 

The other scenario is when we've autoscaling turned on, and in this case, the number of node will be enough to handle the existing kafkas (the Autoscaler will determine) that. 

> by reconciling we may be pushing them to wait for Node provision time along with a rolling update. Maybe we should consider adding min == max nodes first and then adjust the min value later? @machi1990

Right. This could ensure that at least a smooth migration plan from manual to dynamic mode. In my opinion, the last step of the migration plan will be to undo this and make `min` to 0.

> Yes, we talked about a similar approach in staging in general for turning on the auto-scaling as you can alleviate some of the slo concern by gradually dropping the min.

:+1: I was thinking of something slightly different i.e instead of setting min=max which might endup not showing any autoscaling behavior, we could set min=(max -3) at first and decrease `min` gradually over the course of the duration. I'll bring it up to discussion internally.

**Comment by rareddy** [comment]
>The other scenario is when we've autoscaling turned on, and in this case, the number of node will be enough to handle the >existing kafkas (the Autoscaler will determine) that.

Are you saying when the machine pools are automatically created by CP (instead of manual sop) this is in place?


> This could ensure that at least a smooth migration plan from manual to dynamic mode.

hmm, my understanding is in manual mode `spec.capacity` is not defined, so no profiles, there is `kafka` machine pool but not specific `standard` or `developer` machine pools, based on the answer to above comment this is may not be an issue, but we still need a lead time between Node creation and changing the MKA's `spec.capacity` values IMO. Or we could take manual creation of machine pools on existing clusters and reserve the auto-creation to newly created data planes.

**Comment by machi1990** [comment]
> Are you saying when the machine pools are automatically created by CP (instead of manual sop) this is in place?

For manual mode, the machine pool is manually created and we'll need an SOP in place to migrate clusters from the manual list to the dynamic way: add labels, adding taints, make the machine pool autoscaled etc. There is already a JIRA issue created for this. 

What I was saying is, node count won't big of an issue because we'll have what's needed in either scenario. 
What will be an issue is the addition of node taints: there are ways to make the migration smooth which could be addressed by in the SOP. 

> hmm, my understanding is in manual mode spec.capacity is not defined, so no profiles, there is kafka machine pool but not specific standard or developer machine pools

Yes, this is correct. 

> but we still need a lead time between Node creation and changing the MKA's spec.capacity values IMO

I agree.  

> Or we could take manual creation of machine pools on existing clusters and reserve the auto-creation to newly created data planes.

I think manually migrating the existing clusters into the new way of doing things might be easier rather than maintaining the two ways of doing things. The migration steps will be documented in the SOP.

> either way, let's keep documenting these into the release steps pipeline, and when to do a certain task.

+1.

**Comment by shawkins** [comment]
If there are no objections, I'll add another commit here to cover the blast radius expectation.

**Comment by rareddy** [comment]
> If there are no objections, I'll add another commit here to cover the blast radius expectation.

+1

**Comment by shawkins** [comment]
On the blast radius - in retrospect that should have been introduced for the other MGDSTRM-8370 changes as adding the toleration also causes the rolling.  Since that is already in .26, I don't think we need to do anything about it now.

@rareddy  @MikeEdgar The commit looks for a top level dynamic-scaling-scheduling property in the fleetshard_operands.yaml.  It will default to false, so at least a couple versions of the strimzi bundle will need to have this property set once we are ready to migrate.  Once all relevant environments have been migrated we can go back and remove the use of this flag.

**Review by MikeEdgar — APPROVED** [review]

**Comment by shawkins** [comment]
Rebased and pulled out the blast radius commit - this will need refined as described by https://issues.redhat.com/browse/MGDSTRM-9327

**Comment by shawkins** [comment]
This pr now includes https://issues.redhat.com/browse/MGDSTRM-9327 for the blast radius and https://issues.redhat.com/browse/MGDSTRM-8908 that removes the possibility of broker collocation from the ingress replicas.

**Comment by machi1990** [comment]
> Rebased and pulled out the blast radius commit - this will need refined as described by https://issues.redhat.com/browse/MGDSTRM-9327

Has there been a release of strimzi that can allow me to test this? @MikeEdgar @shawkins

**Comment by shawkins** [comment]
Updated to main again, and requested fresh reviews.  The expectations for the code are described in https://docs.google.com/document/d/1jOuEP0GUWAGUERadfglM_thQCM0vVs1p56IJheJuUWE/edit#

In particular the fleetshard behavior does not change until the ManagedKafkaAgent is updated with profile information that should correspond to the creation of a dedicated machine pool.  From there we won't alter scheduling though until the strimzi version configmap has dynamicScalingScheduling set to true.

**Review by machi1990 — APPROVED** [review]

**Comment by machi1990** [comment]
> From there we won't alter scheduling though until the strimzi version configmap has dynamicScalingScheduling set to true.

@shawkins @MikeEdgar @rareddy  Do we current have a Strimzi operator with the `dynamicScalingScheduling` set to `true`? If not, I think that has to be done as well to enable testing, notably for https://issues.redhat.com/browse/MGDSTRM-3363 which is needed to write the migration SOP https://issues.redhat.com/browse/MGDSTRM-6548

**Comment by MikeEdgar** [comment]
> Do we current have a Strimzi operator with the `dynamicScalingScheduling` set to `true`? If not, I think that has to be done as well to enable testing, notably for https://issues.redhat.com/browse/MGDSTRM-3363 which is needed to write the migration SOP https://issues.redhat.com/browse/MGDSTRM-6548

Not yet, but the configuration can be patched in at run time to the existing configmap to simulate what would happen during an upgrade.

File: operator/src/main/java/org/bf2/operator/managers/CapacityManager.java [hunk 9]
```diff
@@ -0,0 +1,330 @@
...
+        if (previous != null) {
+            Resources resources = Serialization.unmarshal(previous, Resources.class);
+            resourceData.computeIfPresent(resources.profile,
+                    (k, v) -> String.valueOf(Integer.parseInt(v) - resources.units));
+            return true;
+        }
+        return false;
+    }
+
+    public synchronized Optional<OperandReadiness> claimResources(ManagedKafka managedKafka, String profile,
```
MikeEdgar, line 206: Can the visibility of this method be reduced? Just to make it clear it's only used locally.

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
```diff
@@ -0,0 +1,330 @@
...
+                    .of(new OperandReadiness(Status.False, Reason.Rejected, "Cluster has insufficient resources"));
+        }
+
+        // update the configmap in a locked manner - this may fail, but we'll retry
+        resourceConfigMap.getData().put(profile, String.valueOf(newTotal));
+        resourceConfigMap.getData().put(getManagedKafkaKey(managedKafka), Serialization.asJson(resources));
+        client.configMaps()
+                .withName(FLEETSHARD_RESOURCES)
+                .lockResourceVersion(resourceConfigMap.getMetadata().getResourceVersion())
+                .replace(resourceConfigMap);
```
MikeEdgar, line 234: If it does fail, is it a runtime exception? I'm wondering if it's something that should be handled that's expected to occur somewhat frequently.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -538,6 +541,17 @@ private void buildIngressController(String name, String domain,
...
+        // if configured for profiles, move the ingress replicas to the default machine pool by adding a dummy toleration
+        if (OperandUtils.shouldProfileLabelsExist(informerManager.getLocalAgent())
+                && overrideManager.migratedToDynamicScalingScheduling()) {
+            builder
+                .editSpec()
+                    .editOrNewNodePlacement()
+                        .addNewToleration("NoSchedule", "kas-fleetshard-ingress", "Exists", null, null)
+                    .endNodePlacement()
+                .endSpec();
+        }
```
MikeEdgar, line 553: I don't understand this part - wouldn't the `NoSchedule` taint on the profile-specific nodes push them off? Is this just to trigger _some_ change to the IngressController CR to re-evaluate the taints?

machi1990:
> > Do we current have a Strimzi operator with the `dynamicScalingScheduling` set to `true`? If not, I think that has to be done as well to enable testing, notably for https://issues.redhat.com/browse/MGDSTRM-3363 which is needed to write the migration SOP https://issues.redhat.com/browse/MGDSTRM-6548
> 
> Not yet, but the configuration can be patched in at run time to the existing configmap to simulate what would happen during an upgrade.

Okay thanks. What's the name of the configmap?

MikeEdgar:
> Okay thanks. What's the name of the configmap?

There will be one per Strimzi deployment in Strimzi's namespace, with the same names as the deployments. Update the `fleetshard_operands.yaml` entry with `dynamic-scaling-scheduling` at the top level.

MikeEdgar (approved):
A few minor suggestions/questions

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
shawkins: Sure, no problem.

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
shawkins: It would fail with a RuntimeException - KubernetesClientException.  It's not expect to occur frequently.  It would be a corner case of a manual modification or another operator pod making a modification.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
shawkins: > wouldn't the NoSchedule taint on the profile-specific nodes push them off?

No, NoSchedule will just prevent new placements.  The existing will remain in place.

> Is this just to trigger some change to the IngressController CR to re-evaluate the taints?

To force the ingress replica pods to reschedule yes, which will force them to the default pool.

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
```diff
@@ -0,0 +1,330 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Responsible for ensuring the resources for a given ManagedKafka are available
+ * prior to allowing the deployment to proceed.
+ */
+@ApplicationScoped
+public class CapacityManager {
```
machi1990, line 40: @shawkins I think we've to exclude reserved mk from remainingUnits calculations.

At the moment if you've a max nodes N, then `maxUnits = N/unitPerNode` , where `unitPerNode` is either `0.1` for developer instance or `3` for standard instance. This is okay. 

And the remainingUnits is calculate against `maxUnits - usedMKCapacityCount`. The `usedMKCapacityCount` at the moment also includes the reserved MK while it should only include real deployment. That's to say that, when we are deploying a reserved instance, we should claim resources. And when we are deleting a reserved instance we shouldn't release resources.
If we do not do so, we end up in a situation where for example with 3 nodes for standard instance. If you prewarm them initially with 1 reserved instance, then remainingUnits becomes `0`, and any subsequent `real` deployment gets rejected due to insufficient resource

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
```diff
@@ -0,0 +1,330 @@
...
+     */
+    Optional<OperandReadiness> validateResources(ManagedKafka managedKafka, String profile, String claimedResources) {
+        Resources resources = Serialization.unmarshal(claimedResources, Resources.class);
+
+        String message = null;
+
+        if (managedKafka.isReserveDeployment()) {
+            message = String.format("The deployment type for %s/%s was flipped to reserved",
+                    managedKafka.getMetadata().getNamespace(), managedKafka.getMetadata().getName());
+        }
```
machi1990, line 251: This message is shown even when the MK instance in argument was a reserved instance.

```
The deployment type for reserved-kafka-standard-1/reserved-kafka-standard-1 was flipped to reserved
```

shawkins: Yes, that did not translate well from some earlier logic merge.  For the purposes of validation we don't actually care if something flips from a real instance to reserved - it will be the same amount either way.  We need either a higher level enforcement, or logic in KafkaCluster to handle this case.

shawkins: After trying to move this handling it became clear that it was quicker to just leave this check but pair it with another one that prevents the claim for reserved managed kafkas.

This does leave us with a corner case situation which is if you have a reserved deployment that you change to a real one - this will currently leave the reserved Deployments in place if they are already created.  We'll address this in that other JIRA about corner cases.

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
```diff
@@ -203,8 +203,11 @@ private boolean releaseResources(Map<String, String> resourceData, String entryK
         return false;
     }
 
-    public synchronized Optional<OperandReadiness> claimResources(ManagedKafka managedKafka, String profile,
+    synchronized Optional<OperandReadiness> claimResources(ManagedKafka managedKafka, String profile,
             ManagedKafkaAgent agent) {
+        if (managedKafka.isReserveDeployment()) {
```
machi1990, line 208: Do we've to do the same for https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/08a05a131c95d401bad591bcc2790c598d404221/operator/src/main/java/org/bf2/operator/managers/CapacityManager.java#L177 when a reserved instance is deleted (removed from the list of MKs)?

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
machi1990: Sounds good. +1, I thought adding handling of reserved instance would suffice to take care of this as well.

operator/src/main/java/org/bf2/operator/managers/CapacityManager.java
shawkins: By preventing the claim in theory a reserved instance should never make it into the list.  Always deleting covers a very unlikely corner case - that someone toggled the delete flag and change the instance from real to reserved at the same time.

machi1990: Right, thanks. I tested this and everything is working as expected for reserved instance.
I'll move along testing real deployments cases.

rareddy (approved):
lgtm

PR #739, opened 2022-06-02
[MGDSTRM-8716] Decode values when reading from master-secret

MikeEdgar:
Also:
- (sync) Check for presence of MK Oauth clientId/secret before adding to master-secret (NPE avoidance)
- (operator) update tests to encode secret values
- (operator) do not assume secret references point to master-secret

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
@@ -43,7 +43,7 @@ public class SecuritySecretManager {
     InformerManager informerManager;
 
     interface SecretSource {
-        Secret apply(ManagedKafka managedKafka, Secret current, Secret masterSecret);
+        Secret apply(ManagedKafka managedKafka, Secret current);
     }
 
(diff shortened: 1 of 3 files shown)


operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
```diff
@@ -184,22 +179,18 @@ public String getServiceAccountPrincipal(ManagedKafka managedKafka, ServiceAccou
...
+        Secret secret;
 
         if (selector != null
                 && selector.getKey() != null
-                && masterSecret != null
-                && masterSecret.getData().containsKey(selector.getKey())) {
-            return masterSecret.getData().get(selector.getKey());
+                && (secret = cachedOrRemoteSecret(managedKafka, selector.getName())) != null
+                && secret.getData().containsKey(selector.getKey())) {
+            return decode(secret.getData().get(selector.getKey()));
```
rareddy, line 193: was looking all over the PR where the double encode was being prevented :)

rareddy (approved):
lgtm

PR #740, opened 2022-06-03
Add kafka_broker_state as an exported metric

SamBarker:
Part of [MGDSTRM-8172](https://issues.redhat.com/browse/MGDSTRM-8172) to expose broker state to CS-SRE & CEE.

### DIFF
--- a/operator/src/main/resources/kafka-metrics.yaml
+++ b/operator/src/main/resources/kafka-metrics.yaml
@@ -297,3 +297,7 @@ data:
         pattern: >-
           kafka.admin.client<type=admin-client-metrics, client-id=(.+)><>(connection-[\w-]*[\w])
         type: GAUGE
+      - name: kafka_broker_state
+        pattern: >-
+          kafka.server<type=KafkaServer, name=BrokerState>
+        type: GAUGE
--- a/operator/src/main/resources/kafka-metrics.yaml
+++ b/operator/src/main/resources/kafka-metrics.yaml
@@ -297,7 +297,10 @@ data:
         pattern: >-
           kafka.admin.client<type=admin-client-metrics, client-id=(.+)><>(connection-[\w-]*[\w])
         type: GAUGE
-      - name: kafka_broker_state
+      - labels:
+          state_ordinal: $1
+        name: kafka_broker_state
         pattern: >-
-          kafka.server<type=KafkaServer, name=BrokerState>
+          kafka.server<type=KafkaServer, name=BrokerState><>Value: (\d+)
         type: GAUGE
+        value: 1


rareddy (approved):
lgtm

grdryn (approved):
LGTM, and I've verified that I get a meaningful result for that object using JMXTool in a broker container:

```
sh-4.4$ ./bin/kafka-run-class.sh kafka.tools.JmxTool --jmx-url service:jmx:rmi:///jndi/rmi://127.0.0.1:9999/jmxrmi --object-name kafka.server:type=KafkaServer,name=BrokerState --one-time true
Trying to connect to JMX url: service:jmx:rmi:///jndi/rmi://127.0.0.1:9999/jmxrmi.
"time","kafka.server:type=KafkaServer,name=BrokerState:Value"
1654262004419,3
```

I do see that this PR is still in "Draft" state, so I won't merge or anything like that (also, maybe that's the reason the integration test run is skipped?).

operator/src/main/resources/kafka-metrics.yaml
```diff
@@ -297,3 +297,7 @@ data:
         pattern: >-
           kafka.admin.client<type=admin-client-metrics, client-id=(.+)><>(connection-[\w-]*[\w])
         type: GAUGE
+      - name: kafka_broker_state
+        pattern: >-
+          kafka.server<type=KafkaServer, name=BrokerState>
```
k-wall, line 302: I'm wonder if it would be more prometheus native to map the mbean's value to a label, and use a static value as 1.  I think that approach will make for more intuitive PromQL queries that work with count/sum functions.  For instance counting all brokers in recovery will be as simple `sum(kafka_broker_state{state_ordinal='2'})`.  WDYT?

https://github.com/prometheus/jmx_exporter#:~:text=milliseconds%20to%20seconds.-,labels,-A%20map%20of

If we take this approach, this will enable us to use a label replacement at ingress to map the values back to meaningful tokens.  We'd do this in https://github.com/bf2fc6cc711aee1a0c2a/observability-resources-mk/blob/main/resources/prometheus/pod_monitors/kafka-metrics.yaml.  This will lead to queries that self-document.

`sum(kafka_broker_state{state='RECOVERYING'})`

SamBarker: I like the idea but I'm struggling to find a way to map the value to a label. 

Have been having issues with getting the metrics exported today, and having lots of issues with failed ZK connections killing pods which really isn't helping...

grdryn: > I like the idea but I'm struggling to find a way to map the value to a label.

In some of the patterns, you'll see a trailing `:`, which I believe is there to prevent capturing the value (and the `:` itself) in the preceding capture group that's used in a label. [Here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/58847fecb57ed808cd24103c95e1e118479adac0/operator/src/main/resources/kafka-metrics.yaml#L60), for example (and it might be the only example, and I hope it's valid). I think that might be part of how to get the value into a label.

SamBarker: Thanks @grdryn, between that and @k-wall's observation about the docs I got it sorted. I think I was being tripped by a white space problem and or getting the angle brackets wrong.

SamBarker:
@k-wall requested changes made.

k-wall (approved):
looks good to me.

PR #741, opened 2022-06-03
Add option for specifying the addon json when installing addon

tplevko:
This gives the user option to specify `--addon-json-config` when installing the addon. In case the config is not provided - the old approach using templates is followed.

This is useful e.g. for `connectors-operator` we test - this addon requires several parameters to be specified, so the predefined template is something we can't use.


tplevko:
Hi @kornys, could you please have a look at this PR? I don't have permissions to assign reviewers.

kornys (approved):

MikeEdgar (approved):
LGTM

PR #742, opened 2022-06-08
feat: general mechanism to allow a container (or init-container) environment to overridden from the OperatorOverride

k-wall:
A general mechanism to allow a container (or init-container) environment to overridden from the `OperatorOverride`.  Currently it is just applied to the canary image.

(The scope of this PR has been broadened.  Originally we wanted a mechanism to allow canary reconcile and connection check interval to be overridden to help investigate [MGDSTRM-8698](https://issues.redhat.com/browse/MGDSTRM-8698) by having the ability to turn off the connection check.

The scope has been extended to include admin server.  I note that strimzi does let let you override envvars for any of its deployments.


k-wall:
> I think this change would be a prime candidate to use `additionalProperties` in `OperandOverride`. For example, the canary could look for a ~`Map<String, EnvVar>`~ `List<EnvVar>` entry under `additionalProperties.get("env")` that could contain any user-provided environment entries and pass them through to the canary container. That leaves a door open to configure individual canary instances rather than having the config apply to the entire fleet.

I like that idea. I think in the cm it would look like:

```
canary:
  image: quay..
  env:
   - name: FOO
      value: bar
```

 but I don't immediately see how we'd persuade Jackson to manufacture EnvVar objects rather (than Map).   What am I missing?

shawkins:
> but I don't immediately see how we'd persuade Jackson to manufacture EnvVar objects rather (than Map). What am I missing?


Use something like Serialization.jsonMapper().convertValue(map, EnvVar[].class) with the map from the env field.

MikeEdgar:
> We could just spin off a separate `List<EnvVar> env` property on the OperandOverride I suppose rather than use the catch-all.

+1

k-wall:
@shawkins @MikeEdgar how does this look?

I needed to make a decision about whether the override feature should apply to *any* environment variable, even those that probably shouldn't be overridden.  I thought about separating a container's env var into a set of those which are reasonable to override and those which are not (`KAFKA_BOOTSTRAP_SERVERS` for instance).    In the end I decided it were simpler to assume that the feature would be used wisely.

operator/src/main/java/org/bf2/operator/operands/Canary.java
```diff
@@ -343,13 +347,34 @@ private List<EnvVar> buildEnvVar(ManagedKafka managedKafka, Deployment current)
...
         envVars.add(new EnvVarBuilder().withName("STATUS_TIME_WINDOW_MS").withValue(String.valueOf(statusTimeWindowMs)).build());
-        return envVars;
+        return applyEnvironmentOverride(envVars, this.overrideManager.getCanaryOverride(managedKafka.getSpec().getVersions().getStrimzi()).getEnv());
     }
 
     private List<ContainerPort> buildContainerPorts() {
         return Collections.singletonList(new ContainerPortBuilder().withName(METRICS_PORT_NAME).withContainerPort(METRICS_PORT).build());
     }
 
+    private List<EnvVar> applyEnvironmentOverride(List<EnvVar> originals, List<EnvVar> overrides) {
```
MikeEdgar, line 357: This might be better placed in `OperandOverride` itself.

MikeEdgar (commented):
This looks good. Just one comment which supports what you said about making this more generally usable for any of the operands.

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
```diff
@@ -41,6 +45,14 @@ public void setImage(String image) {
             this.image = image;
         }
 
+        public List<EnvVar> getEnv() {
+            return env;
+        }
+
+        public void setEnv(List<EnvVar> env) {
```
SamBarker, line 52: I'm wondering if it would be nicer to work with if the api was more declarative:

```suggestion
        public void overriderEnvVar(String name, String value) {
```
and or

```suggestion
        public void overriderEnvVar(EnvVar override) {
```

It would give us a reasonable place to hook in validation for duplicate overrides or to have a allow/deny list of variables (I'm happy with the idea of trusting it will be used wisely) .

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
```diff
@@ -30,6 +32,8 @@ public class OperandOverrideManager {
     public static class OperandOverride {
         public String image;
 
+        public List<EnvVar> env;
```
SamBarker, line 35: I'm related to the thought about a more declarative api I'm also wondering if the overrides would be better as `Map<String, String>` and only converting to `List<EnvVar>` when rendering final result?

operator/src/main/java/org/bf2/operator/operands/Canary.java
k-wall: Done

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
k-wall: At the moment the code supports the possibility of overriding with an environment variable whose source is an EnvVarSource.  That might be a useful option to retain.

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
k-wall: I struggled to get Jackson to accept this.  Feel free to push a commit or guide me with a response.

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
SamBarker: Fine by me.

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
SamBarker: It would probably end up being some nasty combination of `@JsonIgnore` and `@JsonSetter` for a default visible setter. Probably not worth the effort.

k-wall: Ok, let's leave it.

MikeEdgar (approved):

SamBarker (approved):
LGTM

PR #744, opened 2022-06-15
Fixed error when no quota is available and wanted to reuse an already installed cluster

franvila:
Fixing this issue https://issues.redhat.com/browse/MGDSTRM-8836

When we want to use an already created cluster for developing purposes, we get an error every time there is no quota available to create a new one because it is trying to do that. 

Checking first if it already exists is better than trying and wait for the error.


kornys (approved):

MikeEdgar (approved):

franvila:
@kornys @MikeEdgar thanks for the approval. Can any of you merge it, please?

PR #749, opened 2022-06-21
MGDSTRM-8880 workaround a fabric8 issue

shawkins:
it causes intermediate informer cache observations, which can lead to computing the wrong initial state.


rareddy (approved):
lgtm

PR #753, opened 2022-06-22
[MGDSTRM-8880] Avoid NPE when informer event handler is null

MikeEdgar:


shawkins:
Arg, upstream it's tolerant to a null handler on the initial inform calls, but not on the add calls.

rareddy (approved):
lgtm

PR #755, opened 2022-06-23
[MGDSTRM-8940] Fix null check when mapping OAuth/SSO trusted certificate

MikeEdgar:
Fix null check when mapping OAuth/SSO trusted certificate

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
@@ -25,6 +25,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


vbusch (approved):
lgtm

operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
```diff
@@ -288,15 +289,14 @@ private Secret ssoClientSecretFrom(ManagedKafka managedKafka, Secret current) {
     }
 
     private Secret ssoTlsSecretFrom(ManagedKafka managedKafka, Secret current) {
-        if (!isKafkaAuthenticationEnabled(managedKafka)) {
```
shawkins, line 291: Is the isKafkaAuthenticationEnabled check not valid in general?

MikeEdgar: It's valid in other contexts (e.g. building the Kafka listeners), but here it's not enough. Not sure if that's what you're asking.

rareddy (approved):
lgtm

operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java
```diff
@@ -288,15 +289,14 @@ private Secret ssoClientSecretFrom(ManagedKafka managedKafka, Secret current) {
...
-            return null;
-        }
-
-        return buildSecretFrom(ssoTlsSecretName(managedKafka),
-                "Opaque",
-                managedKafka,
-                current,
-                Map.of("keycloak.crt", managedKafka.getSpec().getOauth().getTlsTrustedCertificate()));
+        return Optional.ofNullable(managedKafka.getSpec().getOauth())
+            .map(ManagedKafkaAuthenticationOAuth::getTlsTrustedCertificate)
```
biswassri, line 293: But if we are changing the structure of this secret, don't we also have to change it in places where we're extracting this the TlsCert/key?

rareddy: it is not changing the structure of the secret, it is same. it inlines the isKafkaAuthenticationEnabled call in that first line.

biswassri: Okay, I understand now how it's the same output. Thanks!

PR #760, opened 2022-06-27
add alter config policy into broker config

showuon:
Signed-off-by: Luke Chen <showuon@gmail.com>

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -588,6 +588,7 @@ public class KafkaCluster extends AbstractKafkaCluster {
         if (managedKafka.getSpec().getVersions().compareStrimziVersionTo(Versions.STRIMZI_CLUSTER_OPERATOR_V0_23_0_4) >= 0) {
             // extension to manage the create topic to ensure valid Replication Factor and ISR
             config.put("create.topic.policy.class.name", "io.bf2.kafka.topic.ManagedKafkaCreateTopicPolicy");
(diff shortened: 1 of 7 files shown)


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -588,6 +588,7 @@ private Map<String, Object> buildKafkaConfig(ManagedKafka managedKafka, Kafka cu
         if (managedKafka.getSpec().getVersions().compareStrimziVersionTo(Versions.STRIMZI_CLUSTER_OPERATOR_V0_23_0_4) >= 0) {
             // extension to manage the create topic to ensure valid Replication Factor and ISR
             config.put("create.topic.policy.class.name", "io.bf2.kafka.topic.ManagedKafkaCreateTopicPolicy");
+            config.put("alter.config.policy.class.name", "io.bf2.kafka.config.ManagedKafkaAlterConfigPolicy");
```
SamBarker, line 591: Do we want to hide this behind the feature flag or are you planning to make `io.bf2.kafka.config.ManagedKafkaAlterConfigPolicy` a no-op implementation when the flag is disabled?

showuon: Since we've added feature flag in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/767 and https://github.com/bf2fc6cc711aee1a0c2a/kas-broker-plugins/pull/42 , I think we only need to add the alter config policy config in this PR. Thanks.

showuon:
Working on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/778 . Close it.

PR #767, opened 2022-07-01
MGDSTRM-9050: expose generic mechanism for injecting additional broker configuration from the operand override

k-wall:
Using this mechanism, we can override configuration passed to the broker by populating the `brokerConfig` object on the operator override.  This will have utility for temporary feature flags as well as experimentation in dev.    Note you can remove a broker config key by passing a null value.

```
oc edit cm -n redhat-managed-kafka-operator  strimzi-cluster-operator.v0.26.0-15
```

```
  fleetshard_operands.yaml: |
    canary:
      init:
        image: null
    admin-server:
      image: null
    kafka:
      brokerConfig:
        kas.policy.topic-config.topic-config-policy-enforced: "true" # Inject a new broker config item
        auto.create.topics.enable: ~ # Null causes an existing one to be removed.
```

and showing it works end to end:

```
% oc logs -f kafka-instance-kafka-0 | grep kas.policy.topic-config.topic-config-policy-enforced
kas.policy.topic-config.topic-config-policy-enforced=true
```

### DIFF
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -32,7 +32,7 @@ environment variable.  It also provides an additional configuration item to the 
         kas.policy.topic-config.topic-config-policy-enforced: true
 ```
 
-This can used to override items of configuration.
+This can be used to override items of configuration.
 
 # Logging / Metrics
 
--- a/ADVANCED.md
+++ b/ADVANCED.md
@@ -4,6 +4,38 @@ OLM deployment
 
 Refining operands
 
+# Operand Override
+
+When a kafka instance is being provisioned, fleetshard looks for a configmap named after the requested
+version of strimzi to provide a source of additional configuration.  The configmap must have the label
+`app=strimzi` and must have a name starts with the prefix `strimzi-cluster-operator`.  The config
+map must have a data-item `fleetshard_operands.yaml`.
+
+If such a configmap exists, it may provide:
+
+1. alternative operand image.
+1. environment variables (supported by canary and admin operand).
+1. additional kafka broker configuration options (supported by the kafka operand).
+
+For instance, here's an example override the canary operand image, and passing an additional
+environment variable.  It also provides an additional configuration item to the broker.
+
+```yaml
+ fleetshard_operands.yaml: |
+    canary:
+      image: quay.io/k_wall/strimzi-canary:0.5.0-122333444455555
+      env:
+      - name: FOO
+        value "true"
+    kafka:
+      brokerConfig:
+        kas.policy.topic-config.topic-config-policy-enforced: true
+```
+
+This can used to override items of configuration.
+
+# Logging / Metrics
+
 ## Configure fleetshard operator and synchronizer logging
 
 If you wish to change the logging behavior of a component at runtime, you should create a configmap with an application.properties file in the operator namespace named *component*_logging_config_override.  The application.properties file should contain logging properties of the form quarkus.log.category.*something*.level.
@@ -203,4 +235,6 @@ the resource.  The operator won't overwrite the changes until the digest of the 
 annotation on the target configmap.
 
 Note that after overridding the metric configuration it is currently necessary to roll the kafka or zookeeper pods to have the changes picked up.
+
+
 
--- a/operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
@@ -89,7 +89,7 @@ public class OperandOverrideManager {
     }
     @JsonIgnoreProperties(ignoreUnknown = true)
     public static class Kafka extends OperandOverride {
-        public Map<String, Object> brokerConfig = Map.of();
+        private Map<String, Object> brokerConfig = Map.of();
 
         public Map<String, Object> getBrokerConfig() {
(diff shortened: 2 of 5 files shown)


operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
```diff
@@ -87,6 +87,19 @@ public List<EnvVar> applyEnvironmentTo(List<EnvVar> originals) {
             return List.copyOf(originalsOrderedMap.values());
         }
     }
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public static class Kafka extends OperandOverride {
+        public Map<String, Object> brokerConfig = Map.of();
```
showuon, line 92: Should this change to `private` scope? Otherwise, we don't need getter/setter for `brokerConfig`, right?

showuon (commented):
Thanks for the feature. Overall LGTM. Left a comment. Also, I think we should put the PR description into somewhere, maybe README? wiki? Or...?

operator/src/main/java/org/bf2/operator/managers/OperandOverrideManager.java
k-wall: Yes, jackson will use the getters/setters so public visibility is unnecessary.

k-wall:
> Thanks for the feature. Overall LGTM. Left a comment. Also, I think we should put the PR description into somewhere, maybe README? wiki? Or...?

I added a section to ADVANCED.md

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
```diff
@@ -341,6 +348,11 @@ void testDeveloperManagedKafka() throws IOException {
     @Test
     void testManagedKafkaToKafkaWithCustomConfiguration() throws IOException {
         ManagedKafka mk = exampleManagedKafka("60Gi");
+        Map<String, Object> brokerConfig = new HashMap<>();
+        brokerConfig.put("broker.foo", "bar");
+        brokerConfig.put("auto.create.topics.enable", null);
+        configureMockOverrideManager(mk, brokerConfig);
```
SamBarker, line 354: ooft. The assertion of that is a bit subtle for my taste. 

Given we explicitly set properties here I'd really like to see an explicit test that they are configured. I appreciate they are covered by `diffToExpected` so I'm persuadable that its over kill, but it doesn't seem quite right to me.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -631,6 +632,18 @@ private Map<String, Object> buildKafkaConfig(ManagedKafka managedKafka, Kafka cu
         if (cruiseControlEnabled) {
             config.put("cruise.control.metrics.topic.min.insync.replicas", this.configs.getConfig(managedKafka).cruiseControl.getMetricReporterTopicMinInsyncReplicas());
         }
+
+        // Override broker config from operand override
+        String strimzi = managedKafka.getSpec().getVersions().getStrimzi();
+        Optional.ofNullable(this.overrideManager.getKafkaOverride(strimzi)).map(OperandOverrideManager.Kafka::getBrokerConfig).orElse(Map.of()).forEach((key, value) -> {
```
SamBarker, line 638: Would the `Optional` be better returned by the `overrideManager`? Or does that really mess with Jackson?

SamBarker (approved):
LGTM. 

Happy for this to merge as is, my comments are optional.

ADVANCED.md
```diff
@@ -4,6 +4,38 @@ OLM deployment
...
+      image: quay.io/k_wall/strimzi-canary:0.5.0-122333444455555
+      env:
+      - name: FOO
+        value "true"
+    kafka:
+      brokerConfig:
+        kas.policy.topic-config.topic-config-policy-enforced: true
+```
+
+This can used to override items of configuration.
```
showuon, line 35: nit: This can [be] used to ...

showuon (approved):
Thanks for adding doc in ADVANCED.md. LGTM! Left a minor comment.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: Yes I think so, but I think this should be applied to the other getXXXXOverride methods.  I'd prefer this as a separate refactoring in a separate PR.

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
k-wall: I'm can't say I'm thrilled by these tests myself, but I'm not motivated to refactor them now.  I think what I have done fits the current approach and passes the 'good enough' time.

k-wall:
@MikeEdgar any comments from you?

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
SamBarker: Fine with me

MikeEdgar (approved):
LGTM

PR #769, opened 2022-07-11
haproxy: allow configuration of hard-stop-after configuration and contstats.

k-wall:
This change aims to work around issues on the ingress/egress metrics emitted by the service.

It was discovered that the data transfer metrics (which is exposed to the service user via the console and fleetmanager metrics endpoint) is being reported incorrectly.  In short, this is due to the fact that haproxy [counters used for statistics calculation are incremented only when a session finishes](https://cbonte.github.io/haproxy-dconv/2.2/configuration.html#4.2-option%20contstats) coupled with the way haproxy is reconfigured with new configuration means in progress ingress and egress metrics for inflight connections are lost every time there is a provision or deprovision kafka instance event on the dataplane. For applications like kafka where connections are long lived, this can result in a significantly under estimates for ingress/egress values.

This PR attempts to resolve this issue.  There are three parts:

1. enable haproxy `option contstats` so that haproxy continuous emits statistics for the connection ingress/egress values.  With the old behaviour the stats were emitted when the connection closed, meaning the stat was likely lost if haproxy was restarted,
1. enable haproxy `hard-stop-after` so that when the operator recreates the haproxy process due to a reconfiguration event, the old haprocess dies quickly causing the client to reconnect.  With the old behaviour, the old haproxy remained until the client closed the connection.  This had the downside that the metrics for those connection were never accounted for. 
1. increase `reload-interval` from default of `5s` to `60s` so that successive reconfigurations events are coalesced.   This means there are fewer haproxy restarts and thus fewer client disconnects.  During a kafka instance reprovision there is a reconfiguration event per OpenShift Route (so without this change there are 5 restart events for a 3 broker instance).

This work will hopefully be superseded by https://issues.redhat.com/browse/RFE-3007 and possibly https://issues.redhat.com/browse/NE-879.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -161,6 +161,8 @@ public class IngressControllerManager {
     int maxIngressConnections;
     @ConfigProperty(name = "ingresscontroller.hard-stop-after")
     Optional<String> hardStopAfter;
+    @ConfigProperty(name = "ingresscontroller.ingress-container-command")
+    Optional<List<String>> ingressContainerCommand;
     @ConfigProperty(name = "ingresscontroller.peak-throughput-percentage")
     int peakPercentage;
 
@@ -344,17 +346,19 @@ public class IngressControllerManager {
             log.errorf("Wrong number of containers for Deployment %s/%s", d.getMetadata().getNamespace(), d.getMetadata().getName());
             return false;
         }
-        return !containers.get(0).getResources().equals(deploymentResourceRequirements);
+        Container ingressContainer = containers.get(0);
+        return !(ingressContainer.getResources().equals(deploymentResourceRequirements) && Objects.equals(ingressContainer.getCommand(), ingressContainerCommand.orElse(null)));
     }
 
     private void doIngressPatch(Deployment d) {
-        log.infof("Updating the resource limits for Deployment %s/%s", d.getMetadata().getNamespace(), d.getMetadata().getName());
+        log.infof("Updating the resource limits/container command for Deployment %s/%s", d.getMetadata().getNamespace(), d.getMetadata().getName());
         openShiftClient.apps().deployments().inNamespace(d.getMetadata().getNamespace())
             .withName(d.getMetadata().getName()).edit(
                 new TypedVisitor<ContainerBuilder>() {
                     @Override
                     public void visit(ContainerBuilder element) {
                         element.withResources(deploymentResourceRequirements);
+                        element.withCommand(ingressContainerCommand.orElse(null));
                     }
                 });
     }
@@ -554,7 +558,7 @@ public class IngressControllerManager {
                 - replicationThroughput - throughput / 2 - Quantity.getAmountInBytes(Quantity.parse("1Mi")).longValue();
 
         if (throughputPerIngressReplica < 0) {
-            throw new AssertionError("Cannot appropirately scale ingress as collocating with a broker takes more than the availalbe node bandwidth");
+            throw new AssertionError("Cannot appropriately scale ingress as collocating with a broker takes more than the available node bandwidth");
         }
 
         // average of total ingress/egress in this zone
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -163,6 +163,10 @@ public class IngressControllerManager {
     Optional<String> hardStopAfter;
     @ConfigProperty(name = "ingresscontroller.ingress-container-command")
     Optional<List<String>> ingressContainerCommand;
+    @ConfigProperty(name = "ingresscontroller.reload-interval-seconds")
+    Optional<Long> ingressReloadIntervalSeconds;
+
+
     @ConfigProperty(name = "ingresscontroller.peak-throughput-percentage")
     int peakPercentage;
 
@@ -532,6 +536,14 @@ public class IngressControllerManager {
                     .endNodePlacement()
                 .endSpec();
         }
+
+        ingressReloadIntervalSeconds.ifPresent(value -> {
+            builder.editSpec()
+                    .withNewConfigUnsupportedConfigOverrides()
+                    .withAdditionalProperties(Map.of("reloadInterval", value))
+                    .endConfigUnsupportedConfigOverrides()
+                    .endSpec();
+        });
 
         createOrEdit(builder.build(), existing);
     }
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -24,6 +24,7 @@ import io.fabric8.kubernetes.client.informers.cache.Cache;
 import io.fabric8.kubernetes.client.utils.CachedSingleThreadScheduler;
 import io.fabric8.kubernetes.client.utils.Serialization;
 import io.fabric8.openshift.api.model.Route;
+import io.fabric8.openshift.api.model.operator.v1.Config;
 import io.fabric8.openshift.api.model.operator.v1.IngressController;
 import io.fabric8.openshift.api.model.operator.v1.IngressControllerBuilder;
 import io.fabric8.openshift.api.model.operator.v1.IngressControllerList;
@@ -160,11 +161,11 @@ public class IngressControllerManager {
     @ConfigProperty(name = "ingresscontroller.max-ingress-connections")
     int maxIngressConnections;
     @ConfigProperty(name = "ingresscontroller.hard-stop-after")
-    Optional<String> hardStopAfter;
+    String hardStopAfter;
     @ConfigProperty(name = "ingresscontroller.ingress-container-command")
-    Optional<List<String>> ingressContainerCommand;
+    List<String> ingressContainerCommand;
     @ConfigProperty(name = "ingresscontroller.reload-interval-seconds")
-    Optional<Long> ingressReloadIntervalSeconds;
+    Long ingressReloadIntervalSeconds;
 
 
     @ConfigProperty(name = "ingresscontroller.peak-throughput-percentage")
@@ -351,7 +352,7 @@ public class IngressControllerManager {
             return false;
         }
         Container ingressContainer = containers.get(0);
-        return !(ingressContainer.getResources().equals(deploymentResourceRequirements) && Objects.equals(ingressContainer.getCommand(), ingressContainerCommand.orElse(null)));
+        return !(ingressContainer.getResources().equals(deploymentResourceRequirements) && Objects.equals(ingressContainer.getCommand(), ingressContainerCommand));
     }
 
     private void doIngressPatch(Deployment d) {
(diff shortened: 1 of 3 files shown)


operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -341,17 +350,19 @@ private boolean shouldReconcile(Deployment d) {
...
     private void doIngressPatch(Deployment d) {
-        log.infof("Updating the resource limits for Deployment %s/%s", d.getMetadata().getNamespace(), d.getMetadata().getName());
+        log.infof("Updating the resource limits/container command for Deployment %s/%s", d.getMetadata().getNamespace(), d.getMetadata().getName());
         openShiftClient.apps().deployments().inNamespace(d.getMetadata().getNamespace())
             .withName(d.getMetadata().getName()).edit(
                 new TypedVisitor<ContainerBuilder>() {
                     @Override
                     public void visit(ContainerBuilder element) {
                         element.withResources(deploymentResourceRequirements);
+                        element.withCommand(ingressContainerCommand.orElse(null));
```
shawkins, line 365: Just to make sure, this isn't a part of the deployment that the controller is using to compute its modification hash from right - that is we can modify it and the ingress operator won't immediately try to flip it back.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -158,6 +159,14 @@ public class IngressControllerManager {
     Quantity maxIngressThroughput;
     @ConfigProperty(name = "ingresscontroller.max-ingress-connections")
     int maxIngressConnections;
+    @ConfigProperty(name = "ingresscontroller.hard-stop-after")
+    Optional<String> hardStopAfter;
```
shawkins, line 163: Since these appear in the application.properties, will they ever actually be optional?  I'm fine leaving the code this way, just more of a general question.

shawkins (approved):
LGTM, just a couple of minor comments/questions

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -483,11 +497,15 @@ private void buildIngressController(String name, String domain,
             replicas = Math.max(existingReplicas, replicas);
         }
 
+        Map<String, String> annotations = new HashMap<>();
+        hardStopAfter.ifPresent(value -> annotations.put(HARD_STOP_AFTER_ANNOTATION, value));
```
SamBarker, line 501:
```suggestion
hardStopAfter.ifPresent(value -> builder
                .editOrNewMetadata()
                .withAnnotations(Map.of(HARD_STOP_AFTER_ANNOTATION, value)));
```

I'm not sure if calling `editOrNewMetadata` is safe here, the alternative would be capture the reference to the Metadata on line 505 and use that in the ifPresent lambda. Would avoid any fun with empty vs null

operator/src/main/resources/application.properties
```diff
@@ -38,6 +38,14 @@ ingresscontroller.peak-throughput-percentage=40
 # old sizing - this should go higher, but we should need more memory
 ingresscontroller.max-ingress-connections=75000
 
+# Enables haproxy option contstats in the kas ingress so that stats are reported live rather than at connection close (RFE-3007 provide a proper mechanism to enable this option).
+ingresscontroller.ingress-container-command=/usr/bin/bash,-c,awk '{print $0} /^defaults$/ {print \"  option contstats\"}' < $TEMPLATE_FILE > /tmp/haproxy-config.template; exec /usr/bin/openshift-router --v=2 --template /tmp/haproxy-config.template
```
SamBarker, line 42: 🤠 

And I thought my suggestion of doing it as an image build was pushing the bounds of dirty hacks.  This is next level evil genius stuff.

operator/src/main/resources/application.properties
```diff
@@ -38,6 +38,14 @@ ingresscontroller.peak-throughput-percentage=40
...
+# Disconnect established connections after a haproxy reconfiguration (NE-879 should eliminate the need for this)
+ingresscontroller.hard-stop-after=5s
+# Coalesce up-to reloadInterval worth of haproxy updates.  This prevents a flurry of haproxy restarts (1 per OpenShift Route) as each kafka
+# instance is created (NE-879 should eliminate the need for this).
+ingresscontroller.reload-interval-seconds=60
```
SamBarker, line 47: Was 60s selected based on anything in particular? 

I was going to suggest 30s but thinking a bit further it might be wise to make sure the duration is greater than the period between prometheus scrapes.

I'm just conscious that it will slow down the availability of instance changes. I don't have a good view on how that would play with things like the console/cli saying an instance is available.

SamBarker (approved):
Good work @k-wall. As you say RFE-3007 is a much better fix if we can get it in.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -483,11 +497,15 @@ private void buildIngressController(String name, String domain,
...
+
         builder
             .editOrNewMetadata()
                 .withName(name)
                 .withNamespace(INGRESS_OPERATOR_NAMESPACE)
                 .withLabels(OperandUtils.getDefaultLabels())
+                .withAnnotations(annotations.isEmpty() ? null : annotations)
```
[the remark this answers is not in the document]
LiamClarkeNZ: Why not conditionally apply the `withAnnotations` instead to avoid semantic snares in the underlying library?

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
LiamClarkeNZ: `editOrNewMetadata` is the right approach here, but I'd suggest using `.addToAnnotations(key, val)` to add a single annotation instead, to ensure you preserve any annotations that may be applied previously.

operator/src/main/resources/application.properties
SamBarker: Is this safe in the face of a re-started pod? If the pod is re-started we are going to end up with duplicated `option contstats` entries aren't we?

(Firing up a openshift cluster now to test this)

showuon (approved):
LGTM. Thanks for the patch.

operator/src/main/resources/application.properties
k-wall: The operator creates the openshift routes sequentially, so my aim was to ensure that those routes (n [one for each broker] + 1 bootstrap + 1 admin) have a reasonable chance of being coalesced into a single haproxy restart.  Experiment 30s seemed to be enough for a 3 broker instance.  I didn't test 6.  I doubled it for a safety margin.

During provisioning IIRC Route53 registration will go on in parallel.  In practice, that takes longer than 1 minute, so I don't expect this to impact provision time.  Unfortunately, we can't test this part in development.    Route 53 isn't part of the solution until Stage.

operator/src/main/resources/application.properties
k-wall: > 🤠
> 
> And I thought my suggestion of doing it as an image build was pushing the bounds of dirty hacks. This is next level evil genius stuff.

image

SamBarker: Good point re: ` /tmp/haproxy-config.Template` I hadn't noticed the path change

operator/src/main/resources/application.properties
SamBarker: Cool that seems reasonable for prod,I don't suppose it makes any real difference in practice for kas-installer environments either.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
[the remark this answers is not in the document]
k-wall: Done

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -537,13 +534,26 @@ private void buildIngressController(String name, String domain,
...
+        } else {
+            builder.editMetadata().removeFromAdditionalProperties(HARD_STOP_AFTER_ANNOTATION).endMetadata();
+        }
+
+
+        // intent here is to preserve any other UnsupportedConfigOverrides and just add/remove reloadInterval as necessary.
+        // Surely there a better way to express this?
+        var specNestedConfigUnsupportedConfigOverridesNested = builder.editSpec().withNewConfigUnsupportedConfigOverrides();
+        HasMetadata current = builder.editSpec().buildUnsupportedConfigOverrides();
+        if (current instanceof Config) {
```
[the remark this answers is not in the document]
k-wall: Thanks. I see now https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L247 and https://github.com/fabric8io/kubernetes-client/blob/296ec4fea3e4369c8a248220fb9abcad969ce6d3/kubernetes-model-generator/openshift-model-operator/src/main/resources/schema/kube-schema.json#L3425 is indeed untypes.

So how is it builder.editSpec().withNewConfigUnsupportedConfigOverrides() has sufficient information to return a nested config builder?  What is it that's hinting to the code generation that that is the required type?  Why the lack of symmetry?


Separately - are there any improvements I can make to the code?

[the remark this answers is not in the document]
shawkins: > So how is it builder.editSpec().withNewConfigUnsupportedConfigOverrides() has sufficient information to return a nested config builder?

I take back what I was saying about sundrio not knowing about the sub-types.  It does seem to know about the sub-types in the current module (ImagePruner, KubeControllerManager, Storage, etc.) and to things that have a BuildableReference.  So there are withNewXXXUnsupported... methods for all of those - but that does seem kind of arbitrary given how many possible HasMetadata types there are.

> What is it that's hinting to the code generation that that is the required type? Why the lack of symmetry?

It seems like supporting editOrNewUnsupportedConfigOverrides wouldn't be possible as there are multiple types, but it does seem like it could support editOrNewConfigUnsupportedConfigOverrides - and throw a class cast exception if the existing is not a Config.

> Separately - are there any improvements I can make to the code?

To sundrio?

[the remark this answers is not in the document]
k-wall: > > Separately - are there any improvements I can make to the code?
> 
> To sundrio?

No - I meant to this PR.  Did my last commit leave the code in a reasonable state.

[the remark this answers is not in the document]
shawkins: > I don't understand what brings withNewConfigUnsupportedConfigOverrides into existance right now.

https://github.com/sundrio/sundrio - it's annotations are added to the generated fabric8 object model https://github.com/fabric8io/kubernetes-client/blob/296ec4fea3e4369c8a248220fb9abcad969ce6d3/kubernetes-model-generator/openshift-model-operator/src/generated/java/io/fabric8/openshift/api/model/operator/v1/IngressControllerSpec.java#L62  It must have logic that checks for all the concrete types - but it is only looking in the current module, or to types it has a BuildableReference to.

> No - I meant to this PR. Did my last commit leave the code in a reasonable state?

Yes, it looks good.

PR #771, opened 2022-07-12
[MGDSTRM-9789] Suspend (and resume) Kafka instance based on label

MikeEdgar:
image

- Adding the `bf2.org/suspended=true` label on the `ManagedKafka` CR will scale down all pods running for the instance
- Removing the `bf2.org/suspended` label will scale up the instance pods
- During an upgrade:
  - Status reason remains  'Suspended'
  - Canary scaled up, necessary for stability check
  - Strimzi un-paused, allowed to reconcile resources


operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -76,16 +79,35 @@ public void onUpdate(Kafka oldObj, Kafka newObj) {
...
+        ref.set(object);
+        return ref;
+    }
+
+    public void createOrUpdateMetrics(ManagedKafka managedKafka) {
+        AtomicReference<ManagedKafka> ref = getMetricHolder(managedKafka, managedKafkaMap);
+        Tags tags = MetricsManager.buildKafkaInstanceTags(managedKafka);
 
+        meterRegistry.gauge(MetricsManager.KAFKA_INSTANCE_SUSPENDED, tags, ref, this::suspended);
+        meterRegistry.gauge(MetricsManager.KAFKA_INSTANCE_PAUSED, tags, ref, this::paused);
```
MikeEdgar, line 99: @k-wall , for the new gauge indicating the suspended state of an instance (0 or 1), you had requested the metric always be present. This PR also makes the recently-added metric for the paused state follow the same pattern - always present and a value of 0 or 1. Please let me know if you have any concerns with that approach.

k-wall: @tinaselenge FYI

k-wall: @MikeEdgar I think that sounds good

operator/src/main/java/org/bf2/operator/clients/canary/CanaryPodExecutorService.java
```diff
@@ -0,0 +1,67 @@
...
+                    kubernetesClient
+                        .pods()
+                        .inNamespace(managedKafka.getMetadata().getNamespace())
+                        .withName(canaryPod.getMetadata().getName())
+                        .writingOutput(data)
+                        .writingError(data)
+                        .usingListener(new ExecListener() {
+                            @Override
+                            public void onClose(int code, String reason) {
+                                promise.complete(data.toString(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
MikeEdgar: That's right. This naively assumes we have the payload in `data` by this point in the call. Since this is only for quarkus:dev mode, it should be safe for now.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaController.java
```diff
@@ -214,6 +217,11 @@ private void updateManagedKafkaStatus(ManagedKafka managedKafka, Optional<Operan
             status.setAdminServerURI(kafkaInstance.getAdminServer().uri(managedKafka));
             status.setServiceAccounts(managedKafka.getSpec().getServiceAccounts());
         }
+
+        if (!Objects.equals(originalStatus, status)) {
+            status.setUpdatedTimestamp(ConditionUtils.iso8601Now());
```
shawkins, line 222: @machi1990 a quick search of the fleetmanager code base suggests that the managed kafka status updated timestamp is not being used.  Is this something that could be removed in another JIRA?

MikeEdgar:
Moving back to draft status. I am seeing some flaky behavior during the upgrade process.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaController.java
machi1990: You are correct. We do not use nor have it in the status schema: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/009aca3b4daff751d26bade6827cdac66bf64e0d/openapi/kas-fleet-manager-private.yaml#L403-L470

Was it meant to be used in some cases? I don't seem to recall if we've something lined up that will make use of it so +1 on removing it.

api/src/main/java/org/bf2/operator/ManagedKafkaKeys.java
```diff
@@ -61,7 +61,11 @@ public static final class Annotations {
         /**
          * Set of annotations managed by the data plane - expand as needed.
          */
-        public static final Set<String> DATA_PLANE_ANNOTATIONS = Set.of(MASTER_SECRET_DIGEST, PAUSE_RECONCILIATION);
+        public static final Set<String> DATA_PLANE_ANNOTATIONS = Set.of(
+                MASTER_SECRET_DIGEST,
+                KAFKA_UPGRADE_START_TIMESTAMP,
+                KAFKA_UPGRADE_END_TIMESTAMP,
```
MikeEdgar, line 67: It's not clear to me how having these annotations missing today does not interfere with the Kafka version upgrade process.

shawkins: It sure seems like it could be a problem.  Should this be separated into a different pr for immediate inclusion?

MikeEdgar: After discussing in the standup, I think we're safe to leave this as-is without an immediate/separate fix.

operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
```diff
@@ -227,6 +262,12 @@ public void createOrUpdate(ManagedKafka managedKafka) {
         Kafka current = cachedKafka(managedKafka);
         Kafka kafka = kafkaFrom(managedKafka, current);
         createOrUpdate(kafka);
+
+        boolean reconciliationPaused = isReconciliationPaused(managedKafka);
+
+        if (managedKafka.isSuspended() && reconciliationPaused && !updatesInProgress(managedKafka)) {
```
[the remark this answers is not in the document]
MikeEdgar: This is necessary so that we can remove the pods without Strimzi creating them again.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -1112,7 +1140,14 @@ private Map<String, String> buildKafkaAnnotations(ManagedKafka managedKafka, Kaf
         if (annotations == null) {
             annotations = new HashMap<>();
         }
-        //this.strimziManager.togglePauseReconciliation(managedKafka, this, annotations);
+
+        if (managedKafka.isSuspended() && !blockSuspension(managedKafka)) {
+            annotations.put(StrimziManager.STRIMZI_PAUSE_RECONCILE_ANNOTATION, "true");
+            annotations.remove(ManagedKafkaKeys.Annotations.STRIMZI_PAUSE_REASON);
+        } else if (!annotations.containsKey(ManagedKafkaKeys.Annotations.STRIMZI_PAUSE_REASON)) {
+            annotations.remove(StrimziManager.STRIMZI_PAUSE_RECONCILE_ANNOTATION);
```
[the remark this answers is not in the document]
MikeEdgar: The pause reason is an MK annotation, not Strimzi.

rareddy (approved):
lgtm

I did deploy this using kas-installer and tested on dev cluster. Reviewed most classes, but some of test classes I didn't follow completely

MikeEdgar:
@machi1990 , @k-wall - if you have a chance, please take a look (or delegate) from the control plane and Kafka Integrations perspectives.

machi1990:
> @machi1990 , @k-wall - if you have a chance, please take a look (or delegate) from the control plane and Kafka Integrations perspectives.

@pawelpaszki @JameelB can one of you have a look from the control plane perspective? Thanks

MikeEdgar:
@shawkins , please take another look at this one if you have time.

operator/src/main/java/org/bf2/operator/managers/KafkaManager.java
```diff
@@ -428,12 +429,16 @@ private boolean isKafkaUpgradeStabilityCheckJobExists(ManagedKafka managedKafka)
...
         log.debugf("[%s/%s] Adding Kafka upgrade %s timestamp annotation",
                 managedKafka.getMetadata().getNamespace(), managedKafka.getMetadata().getName(), annotation);
+
+        String timestampValue = ZonedDateTime.now(ZoneOffset.UTC).format(DateTimeFormatter.ISO_INSTANT);
+        managedKafka.getMetadata().getAnnotations().put(annotation, timestampValue);
+
         managedKafkaClient
                 .inNamespace(managedKafka.getMetadata().getNamespace())
                 .withName(managedKafka.getMetadata().getName())
                 .edit(mk -> new ManagedKafkaBuilder(mk)
```
shawkins, line 439: based upon the modification the managedKafka above, this could just be .edit(mk -> managedKafka)

MikeEdgar: @shawkins , I was actually questioning whether updating the MK in this class is the right thing to do. I believe it frequently results in HTTP/409 when the operator SDK goes to update the resource after the reconcile. I wouldn't want to do anything about that in this PR, but maybe it deserves some attention at some point.

shawkins: Yes, if this is happening in-line with the main reconciliation loop, then it's definitely wrong to issue this additional update.  Instead ManagedKafkaController should return UpdateControl.updateResourceAndStatus.

operator/src/main/java/org/bf2/operator/managers/MetricsManager.java
```diff
@@ -76,16 +79,35 @@ public void onUpdate(Kafka oldObj, Kafka newObj) {
...
-        ref.set(kafka);
+        ref.set(object);
+        return ref;
+    }
+
+    public void createOrUpdateMetrics(ManagedKafka managedKafka) {
+        AtomicReference<ManagedKafka> ref = getMetricHolder(managedKafka, managedKafkaMap);
+        Tags tags = MetricsManager.buildKafkaInstanceTags(managedKafka);
 
+        meterRegistry.gauge(MetricsManager.KAFKA_INSTANCE_SUSPENDED, tags, ref, this::suspended);
```
JameelB, line 98: @pawelpaszki note that this PR adds the `kafka_instance_suspended` metric. Seems like this will always be present whether the Kafka instance is suspended (1) or not (0)

JameelB (approved):
lgtm on control plane perspective. Status reported by fleetshard is as expected. I believe this has also been tested with the control plane changes by Mike.

operator/src/main/java/org/bf2/operator/managers/KafkaManager.java
MikeEdgar: Opened MGDSTRM-10094 to track this.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaController.java
MikeEdgar: Tracking the removal of the timestamp in MGDSTRM-10096

PR #772, opened 2022-07-14
MGDSTRM-8991 making sure the observability secret has an owner

shawkins:
I'm not sure why we got into the state observed on MGDSTRM-8991 - something removed the owner reference.  When that happens the existing edit/update logic will not add it back.  So this set of modifications makes the logic more consistent with other places.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/ObservabilityManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/ObservabilityManager.java
@@ -17,6 +17,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 @ApplicationScoped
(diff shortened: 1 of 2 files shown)


biswassri:
@shawkins no problem Steve, I hadn't started working on this issue. I'll reassign the issue to you if anyone comebacks for clarity.

rareddy (approved):
lgtm

shawkins:
> Just to mention it, the message Secret redhat-kas-fleetshard-operator/fleetshard-observability does not have OwnerReference is logged the moment the ownerReference is removed from the secret prior to being added back by the fix in this PR.

The implication is that the observability operator must be removing the ownerReference as there's no path in the fleetshard that will remove it.  At least with this change we'll be good on our end with the expected state of the secret. @pb82 do you want another JIRA to cover looking at this from the observability side?

PR #775, opened 2022-07-18
MGDSTRM-8585 adding support for reserved deployments

shawkins:
Updating the pr to be independent of capacity reporting.  If the operator sees a deployment marked as reserved, then it won't actually create the expected resources.  Instead deployments with pause pods at a lower priority will be created with resources and replicas that match the expected instances.  The parent managedkafka will be marked as ready once all of these deployments are ready.

The priority class kas-fleetshard-reservation (which I'm open to changing) was added to the kubernetes.yml in the project as we need it to be part of the workload, which diverges from the kas-fleetshard-high used in the bundle for the operator components.

There is no interaction with metrics as those are derived from kafka resources.

This change has no expectation that profile specific machine pools will be in use.

I'll run this on my local cluster again to make sure I didn't miss anything pulling this over to main.

### DIFF
--- a/operator/src/main/kubernetes/kubernetes.yml
+++ b/operator/src/main/kubernetes/kubernetes.yml
@@ -231,3 +231,10 @@ spec:
     matchLabels:
       app: kas-fleetshard-operator
 ---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: kas-fleetshard-reservation
+value: -4
+globalDefault: false
+---
--- a/api/src/main/java/org/bf2/operator/resources/v1alpha1/ManagedKafka.java
+++ b/api/src/main/java/org/bf2/operator/resources/v1alpha1/ManagedKafka.java
@@ -27,6 +27,8 @@ import java.util.UUID;
 @Version("v1alpha1")
 public class ManagedKafka extends CustomResource<ManagedKafkaSpec, ManagedKafkaStatus> implements Namespaced {
 
+    public static final String RESERVED_DEPLOYMENT_TYPE = "reserved";
+
     private static final String CERT = "cert";
 
     public static final String BF2_DOMAIN = "bf2.org/";
@@ -35,6 +37,8 @@ public class ManagedKafka extends CustomResource<ManagedKafkaSpec, ManagedKafkaS
 
     public static final String PROFILE_TYPE = BF2_DOMAIN + "kafkaInstanceProfileType";
     public static final String PROFILE_QUOTA_CONSUMED = BF2_DOMAIN + "kafkaInstanceProfileQuotaConsumed";
+
+    public static final String DEPLOYMENT_TYPE = BF2_DOMAIN + "deployment";
 
     @Override
     protected ManagedKafkaSpec initSpec() {
@@ -172,6 +176,14 @@ public class ManagedKafka extends CustomResource<ManagedKafkaSpec, ManagedKafkaS
     public static ManagedKafka getDummyInstance(int name) {
         return getDefault("mk-" + name, "mk-" + name, "xyz.com", CERT, CERT, "clientId", CERT, "secret",
                 "claim", "fallbackClaim", "http://jwks", "https://token", "http://issuer", "strimzi-cluster-operator.v0.23.0", "2.7.0");
+    }
+
+    @JsonIgnore
+    public boolean isReserveDeployment() {
+        if (getMetadata().getLabels() == null) {
+            return false;
+        }
+        return RESERVED_DEPLOYMENT_TYPE.equals(getMetadata().getLabels().get(DEPLOYMENT_TYPE));
     }
 
 }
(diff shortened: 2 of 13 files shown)


shawkins:
Checked locally again, this seems to look good.  For a developer instance you see the creation of deployments:

name-admin-server
name-canary
name-cruisecontrol
name-exporter
name-kafka
name-zookeeper

Each creating just a single replica of a pod that has resources matching the real thing.

rareddy:
I checked with my "kas-installer" based install modifying the sync to mimic the control plane to add the label, and I verified that all the needed deployments are created in `reserved` mode, and they have relinquished the space when a real instance needed the resources.

Only one code change request I have is instead of "pausepod" or "pausedeployment" can we generally use "reservedpod" or "reserveddeployment" and leave the technical name of "pause pod" just to the image name?

shawkins:
> Only one code change request I have is instead of "pausepod" or "pausedeployment" can we generally use "reservedpod" or "reserveddeployment" and leave the technical name of "pause pod" just to the image name?

Done

machi1990 (approved):
Checked the contract between the DP and CP, and it looks good. Thanks @shawkins 

@rareddy is there an operator bundle that we can test with what we've on CP side?

rareddy (approved):
lgtm

rareddy:
@machi1990 you can use kas-installer, in there in the operators directory there is script to generate the bundle, you also need to override the ENV property KAS_FLEETSHARD_OLM_INDEX_IMAGE

machi1990:
> @machi1990 you can use kas-installer, in there in the operators directory there is script to generate the bundle, you also need to override the ENV property KAS_FLEETSHARD_OLM_INDEX_IMAGE

Perfect, thanks. I didn't know if we have the ability to generate the bundle in kas-installer.

operator/src/main/kubernetes/kubernetes.yml
```diff
@@ -231,3 +231,10 @@ spec:
     matchLabels:
       app: kas-fleetshard-operator
 ---
+apiVersion: scheduling.k8s.io/v1
```
rareddy, line 234: This does not seem to be the way the priority class is wired to be picked by OLM bundles. I think we need to change bundle scripts and add other priority classes here (probably better) vs define this priority class directly in the bundle scripts of build pipeline as well as kas-installer. For my testing created this manually. but when I tried to with kas-installer I saw it created previous one in the script not from kubernetes.yml @shawkins  @MikeEdgar what do u suggest?

MikeEdgar: The bundle scripts can be updated to expect it in `kubernetes.yml` and use that. We need both included in the bundle correct?

MikeEdgar: We can just expand on what is done here to apply for `PriorityClass` resources.

https://github.com/bf2fc6cc711aee1a0c2a/kas-installer/blob/f9a5fcd65c07174adf6dd0cdf6c88bf688f39ea0/operators/generate-kas-fleetshard-olm-bundle.sh#L182-L184

rareddy: yeah, I think expanding the `kubernetes.yaml` and fixing the bundle scripts is the right approach if at all we need to deal with this again.

shawkins: Yes, I called this out in the initial pr description.  Unlike the other priority class this one is used internally, so it does logically belong in the kubernetes.yml.

machi1990 (approved):
I've tested this with the control plane dev branch on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/1202, other than the mentioned priority class issue which I've had to create manually, everything else worked as expected.

What are the timelines for this one being merged and released? @shawkins @rareddy @MikeEdgar

shawkins:
> What are the timelines for this one being merged and released?

@machi1990 Mike has worked the bundling issue for the priority class, so I think we'll merge this soon.  Do you want to make a further determination on the timeout issue you mentioned in chat first?  Assuming that we're good this should go into .27, but I don't know a date for that.

machi1990:
> > What are the timelines for this one being merged and released?
> 

> Mike has worked the bundling issue for the priority class, so I think we'll merge this soon. 
> Assuming that we're good this should go into .27, but I don't know a date for that.

Thanks. 

> @machi1990 Do you want to make a further determination on the timeout issue you mentioned in chat first? 

Thanks. I've replied via chat, and I do not also think the issue is because of reserved capacity but more generally could be because of autoscaling (node took longer than 420000ms to come up for ZK to be provisioned). So the issue doesn't have to hold this PR. I'll try to reproduce the scenario and share the exact message with you.

shawkins:
If there are no objections, I'll add another commit here to cover the blast radius expectation.

MikeEdgar:
> If there are no objections, I'll add another commit here to cover the blast radius expectation.

I don't object, but isn't this change pretty well contained already? I.e. deploying the operator with this wouldn't cause rolling pods, etc.

shawkins:
> I don't object, but isn't this change pretty well contained already? I.e. deploying the operator with this wouldn't cause rolling pods, etc.

Sorry, I put this comment on the wrong pr.  It should be on the other one that is enforcing the max.

MikeEdgar (approved):

PR #776, opened 2022-07-19
fix: spec/unsupportedConfigOverrides/reloadInterval backing var has wrong type

k-wall:
prevents periodic, believed harmless, messages in the fleetshard logs

```
Updating the existing IngressController openshift-ingress-operator/kas [{"op":"replace","path":"/spec/unsupportedConfigOverrides/reloadInterval","value":60}]
```


MikeEdgar (approved):
LGTM

shawkins (approved):
LGTM

PR #778, opened 2022-07-22
MGDSTRM-9192: add custom topic config policy settings

showuon:
In this PR, I added the following broker configs:
1. `alter.config.policy.class.name` -> "io.bf2.kafka.config.ManagedKafkaAlterConfigPolicy"
2. `kas.policy.topic-config.enforced` -> "flush.ms:9223372036854775807,index.interval.bytes:4096,compression.type:producer,flush.messages:9223372036854775807,min.cleanable.dirty.ratio:0.5,file.delete.delay.ms:60000,segment.index.bytes:10485760,preallocate:false,unclean.leader.election.enable:false" + **min.insync.replicas: {{CURRENT_MIN_INSYNC_SETTING}}** ([MGDSTRM-9202](https://issues.redhat.com/browse/MGDSTRM-9202))
3. `kas.policy.topic-config.range` -> "max.message.bytes::1048588,segment.bytes:52428800:,segment.ms:600000:"
4. `kas.policy.topic-config.mutable` -> "message.timestamp.difference.max.ms,message.timestamp.type,retention.bytes,min.compaction.lag.ms,cleanup.policy,max.compaction.lag.ms,delete.retention.ms,message.downconversion.enable"
5. `kas.policy.topic-config.topic-config-policy-enforced` -> configurable `topic-config-policy-enforced` value, default to `true`
6. `segment.jitter.ms` will be an immutable config ([MGDSTRM-9202](https://issues.redhat.com/browse/MGDSTRM-9202))


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -125,6 +125,32 @@ public class KafkaCluster extends AbstractKafkaCluster {
     public static final String MAX_PARTITIONS = "max.partitions";
     public static final String MESSAGE_MAX_BYTES = "message.max.bytes";
 
+    private static final String ENFORCED_RULE_CONFIG = String.join(",",
```
k-wall, line 128: I don't think we want this hardcoded in the Java code.  You should follow the pattern of putting it in the application.properties (common to all), standard.properties (used by standard( and developer.properties (used by developer).

That will also allow us to to override the settings from environment variables, should the need ever arise.

showuon: Thanks @k-wall for the comments. I'm working on it, and has a question:
After we move the setting to `strandard.properties` and `developer.properties`. The reason is we'll set the enforce rule for standard to `min.insync.replicas:2`, but there is a feature allows user to scale down the cluster, so the `min.insync.replicas` will change to 1.  But the enforce rule doesn't change, so it'll fail.

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/0.21.x/operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java#L257-L260

Since we'll dynamically set `min.insync.replicas` value based on the `scaling-and-replication-factor` property, I think we should also apply the same pattern to `enforce rule`. That is, in enforce rule config, we don't set `min.insync.replicas`, ex:
`kafka.topic-config-enforced-rule="index.interval.bytes:4096,compression.type:producer"`
And during setting the kakfa broker config, we dynamically append "expected min.insync.replicas" value to it, just like how we set kafka broker `min.insync.replicas` value.


WDYT?
I've updated the PR with this solution. Let me know if you have other suggestion. Thank you.

operator/src/main/resources/application.properties
```diff
@@ -142,3 +142,8 @@ group=*;apis=offset_fetch,offset_commit,heartbeat,describe_groups;level=DEBUG
 
 managedkafka.canary.init-enabled=true
 managedkafka.canary.init-timeout-seconds=600
+
+managedkafak.kafka.topic-config-policy-enforced=true
+managedkafak.kafka.topic-config-range-rule=max.message.bytes::1048588,segment.bytes:52428800:,segment.ms:600000:
+managedkafak.kafka.topic-config-mutable-rule=message.timestamp.difference.max.ms,message.timestamp.type,retention.bytes,min.compaction.lag.ms,cleanup.policy,max.compaction.lag.ms,delete.retention.ms,message.downconversion.enable
+managedkafak.kafka.topic-config-enforced-rule=flush.ms:9223372036854775807,index.interval.bytes:4096,compression.type:producer,flush.messages:9223372036854775807,min.cleanable.dirty.ratio:0.5,file.delete.delay.ms:60000,segment.index.bytes:10485760,preallocate:false,unclean.leader.election.enable:false
```
showuon, line 149: set the enforced rule without `min.insync.replias`, and we append it dynamically.

k-wall: > After we move the setting to strandard.properties and developer.properties, there will be a test failed. The reason is we'll set the enforce rule for standard to min.insync.replicas:2, but there is a feature allows user to scale down the cluster, so the min.insync.replicas will change to 1. 

I'm not sure what you mean.  There' no feature that lets a user go from developer to standard or standard to developer.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -628,24 +603,29 @@ private Map<String, Object> buildKafkaConfig(ManagedKafka managedKafka, Kafka cu
...
-        config.put("kas.policy.topic-config.range", RANGE_RULE_CONFIG);
-        config.put("kas.policy.topic-config.mutable", MUTABLE_RULE_CONFIG);
+                kafkaConfigs.isPartitionLimitEnforced());
+
+        boolean isTopicConfigPolicyEnforced = kafkaConfigs.isTopicConfigPolicyEnforced();
+        config.put("kas.policy.topic-config.topic-config-policy-enforced", isTopicConfigPolicyEnforced);
+        if (isTopicConfigPolicyEnforced) {
+            String topicConfigEnforcedRule = kafkaConfigs.getTopicConfigEnforcedRule();
+            if (topicConfigEnforcedRule.contains("min.insync.replicas")) {
+                topicConfigEnforcedRule = topicConfigEnforcedRule.replaceAll("min\\.insync\\.replicas:[0-9]+", "min.insync.replicas:" + minIsr);
```
k-wall, line 615: I don't think this approach is too nice.

operator/src/main/resources/application.properties
```diff
@@ -142,3 +142,8 @@ group=*;apis=offset_fetch,offset_commit,heartbeat,describe_groups;level=DEBUG
 
 managedkafka.canary.init-enabled=true
 managedkafka.canary.init-timeout-seconds=600
+
+managedkafak.kafka.topic-config-policy-enforced=true
```
k-wall, line 146: typo in `managedkafak`

operator/src/main/resources/application.properties
k-wall: It would have been nice if we could have written the topic-config-enforced-rule in terms of existing variable ${kafka.scaling-and-replication-factor}.

developer.properties:
```properties
kafka.scaling-and-replication-factor=1
```

standard.properties:
```properties
kafka.scaling-and-replication-factor=3
```

application.properties:
```properties
managedkafak.kafka.topic-config-enforced-rule=.......,min.insync.replicas:${kafka.scaling-and-replication-factor}
```

Alas, I'm pretty sure SmallRye won't resolve this.  I didn't try btw.  

As an alternative, perhaps pushing the rules down into standard/developer.properties is least worst. You can then reference the ${kafka.scaling-and-replication-factor} (because it is in the same file).  It does have the disadvantage that the config would be duplicated in the property files.

developer.properties:
```properties
managedkafak.kafka.topic-config-enforced-rule=.......,min.insync.replicas:${kafka.scaling-and-replication-factor}
```

Finally perhaps using [SmallRye indexed properties ](https://quarkus.io/guides/config-reference#indexed-properties) would make the config easier to read. Those long lines are pretty hard on the eye. WDYT?   You need to change the property into List, then join with a comma before passing to Kafka.

developer.properties:
```properties
managedkafka.kafka.topic-config-range-rule[0]=max.message.bytes::1048588
managedkafka.kafka.topic-config-range-rule[1]=segment.bytes:52428800:,
managedkafka.kafka.topic-config-range-rule[2]=segment.ms:600000:
```

k-wall:
Maybe closing #760 would be cleanest, and do  everything in this one PR.

operator/src/main/resources/application.properties
showuon: @k-wall , thanks for the suggestion. But I'm afraid it won't fix this issue. If you check this test clearly, you'll find it loads the standard properties file first, and then replace `scaling-and-replication-factor` value next via `setScalingAndReplicationFactor(1)`. That means, no matter what we did in the "property file" level, it'll be replaced in the run time in the end.

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/0.21.x/operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java#L257-L260

So, I think we can only do the same thing as what we did for `min.insync.replicas` config:
https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/main/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java#L649-L654

WDYT?

k-wall:
As this SOP update relies on RPMs being added to the containers, the merging of this SOP be blocked until the next fleetshard/rhosak releases are made.   @StevenTobin FYI

operator/src/main/resources/application.properties
showuon: After discussion, we agreed we can set `TopicConfigPolicyEnforced` to false, to bypass this test. We believe this scaling down/up case won't happen in current production environment.

showuon:
@k-wall , thanks for the comment. I've updated the PR. Some response to your previous suggestion below:

> It would have been nice if we could have written the topic-config-enforced-rule in terms of existing variable ${kafka.scaling-and-replication-factor}.

I've read the doc, and unfortunately, it only works for `application.properties` files (or any file displayed in below image)
image

The `standard.properties` or `developer.properties` files are loaded by ourselves (not via quarkus), so it won't do the expression expansion out-of-box (unless we add this feature by ourselves, but I don't see any value to add it). So I put the enforced rule value hard-coded, standard to 2, and developer to 1.

> Finally perhaps using [SmallRye indexed properties ](https://quarkus.io/guides/config-reference#indexed-properties)would make the config easier to read. Those long lines are pretty hard on the eye. WDYT?

Yes, that makes the value more readable (Interesting feature). But I checked the doc, when reading the config value, it'll convert it to `List<String>`, not pure string as we expected. So I keep it as is to make it simple. 

Let me know if you have any comments. Thank you.

operator/src/main/resources/instances/developer.properties
```diff
@@ -26,5 +26,6 @@ storage.min-margin=1Gi
 # this is arbitrary, we don't expect request for multiple developer units
 kafka.partition-capacity=1500
 kafka.message-max-bytes=1048588
+kafka.topic-config-enforced-rule=flush.ms:9223372036854775807,index.interval.bytes:4096,compression.type:producer,flush.messages:9223372036854775807,min.cleanable.dirty.ratio:0.5,file.delete.delay.ms:60000,segment.index.bytes:10485760,preallocate:false,unclean.leader.election.enable:false,min.insync.replicas:1
```
MikeEdgar, line 29: It seems a little off that the plugin needs the value of `min.insinc.replicas` specified twice. What if the `:1`/value is optional here and the plugin interprets that as "use the config value of `min.insync.replicas` as the enforcement value instead".

showuon: >  What if the `:1`/value is optional here and the plugin interprets that as "use the config value of `min.insync.replicas` as the enforcement value instead".

Sounds good to me. Thanks.
@k-wall , if you have no objection, I'll go with it.

k-wall, line 29:
```suggestion
kafka.topic-config-enforced-rule=flush.ms:9223372036854775807,index.interval.bytes:4096,compression.type:producer,flush.messages:9223372036854775807,min.cleanable.dirty.ratio:0.5,file.delete.delay.ms:60000,segment.index.bytes:10485760,preallocate:false,unclean.leader.election.enable:false,min.insync.replicas:${kafka.scaling-and-replication-factor}
```

k-wall: I made an alternative suggestion to avoid the duplication.

k-wall:
with the resolution of the comment above, lgtm

operator/src/main/resources/instances/developer.properties
showuon: @k-wall , thanks for the comment. But you might want to check my previous comment: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/778#issuecomment-1199130100 . Simply put, the expression expansion (i.e. `${kafka.scaling-and-replication-factor}`) won't work in [developer|standard].properties. It only works in `application.properties`. In this case, I think we can keep it as is. WDYT?

k-wall: @showuon ok - I appreciated that the property refs wouldn't work across files, but failed to appreciate that it applied with the [developer|standard].properties files too.  Thanks for putting me right.

k-wall (approved):

k-wall:
please resolve the threads / squash the changes

showuon:
@MikeEdgar , do you have any other comment? We think we should have less runtime check logic, and just rely on the provided configs. We can add those logic if needed in the future. Thanks.

MikeEdgar (approved):
lgtm. Is #786 the same?

showuon:
Sorry, #786  is a mistake. I've closed it. Thanks.

PR #782, opened 2022-08-01
MGDSTRM-9334 correcting the defaults for developer instances

shawkins:
For now the capacity values are technically optional, so we should have valid defaults - however the control plane is supplying these, so there should be no impact to actual instances.  Once we address https://issues.redhat.com/browse/MGDSTRM-8648 we could remove the possibility of not setting these capacity values.


MikeEdgar (approved):
lgtm

rareddy (approved):
lgtm

PR #785, opened 2022-08-03
Fix for sonar target paths

biswassri:


.github/workflows/build.yml
```diff
@@ -42,8 +42,5 @@ jobs:
       with:
         name: target
         path: |
-          target/
-          api/target/
-          common/target/
-          operator/target/
-          sync/target/
+          **/target/
+          !**/target/**/.jar
```
MikeEdgar, line 46:
```suggestion
          !**/target/**/*.jar
```

MikeEdgar (approved):

MikeEdgar:
This adds the target dirs for all modules and excludes the Quarkus jars from the archive, making it a much more reasonable size.

rareddy (approved):
lgtm

PR #789, opened 2022-08-05
MGDSTRM-9230 removing the drain cleaner pdb for a single replica

shawkins:


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -479,7 +479,7 @@ private KafkaClusterTemplate buildKafkaTemplate(ManagedKafka managedKafka) {
         KafkaClusterTemplateBuilder templateBuilder = new KafkaClusterTemplateBuilder()
                 .withPod(podTemplateBuilder.build());
 
-        if (drainCleanerManager.isDrainCleanerWebhookFound()) {
+        if (replicas > 1 && drainCleanerManager.isDrainCleanerWebhookFound()) {
             templateBuilder.withPodDisruptionBudget(
                 new PodDisruptionBudgetTemplateBuilder()
                     .withMaxUnavailable(0)
```
rareddy, line 485: More of a question, 
1) why is this enabled only when drain cleaner webhook is found?
2) with `maxUnavilable` of `0` it says all the broker nodes must be available. I wonder how the OCP upgrades are taking place, I thought should this be the same as insync-replica count?

MikeEdgar: Testing my understanding of this :) ... having it set to `0` means that we rely on the presence of the drain cleaner to label the pods for Strimzi to restart the pods.

MikeEdgar (approved):
lgtm. Tested it and the PDB `maxUnavailable` changes to `1` and the statuses change to the below condition, which makes sense.

```yaml
  conditions:
    - type: DisruptionAllowed
      status: 'True'
      observedGeneration: 2
      lastTransitionTime: '2022-08-05T17:02:05Z'
      reason: SufficientPods
      message: ''
```

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
shawkins: @MikeEdgar yes that is correct. 

@rareddy see https://github.com/strimzi/drain-cleaner#see-it-in-action

rareddy: Thanks for explanation

rareddy (approved):
Lgtm

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
fvaleri: We set `PDB=0` for Kafka and ZK pods in order to block K8s draining and delegate this task to the combined work of the Drain Cleaner and the Cluster Operator, **regardless of the number of replicas**. This is to avoid service disruption, as the CO knows when a replica is fully synced and we can drain the next one. 

@shawkins this change should be also applied to `buildZookeeperTemplate`, but I think it is not needed. There are corner cases in which it is never safe to roll a broker. A single broker cluster is one, a topic with `RF < min.insync.replicas` is another. The CO is smart enough to recognize these cases and proceed with the broker roll. This is also confirmed by a quick test that I did using the latest version of the DC and the CO.

shawkins: @fvaleri When you say that it's not needed, do you mean that the auto-scaler will still be able to reschedule developer instances away from nodes that are under their utilization threshold - https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#does-ca-work-with-poddisruptionbudget-in-scale-down

Because these are developer instances cost concerns trump safety.

fvaleri: Not exactly, I mean that both the drain command and the cluster-autoscaler will start retrying periodically because of the PDB settings. In the meantime, DC+CO will do their work moving out any ZK and Kafka pod from the terminating node. This means the node drain will eventually succeed. For the drain command the default timeout is infinite, while it looks like the CA waits at most 2 minutes, which is probably not be enough and should be raised based on testing.

shawkins: My reading of the autoscaler source is that prior to attempting the draining, it will do a quick check of if the pdbs are satiable:

https://github.com/kubernetes/autoscaler/blob/5745044ddf4378c1aae16c47fb173b1abf709c8e/cluster-autoscaler/simulator/drain.go#L102

So I think just having the pdb on the developer instances is enough to block the auto-scaling attempt at draining.

> while it looks like the CA waits at most 2 minutes, which is probably not be enough and should be raised based on testing.

Unfortunately we have almost no settings that can be tuned via the openshift autoscaling - for example see https://issues.redhat.com/browse/HIVE-1976 - if a timeout change is needed, it will require a change pretty far upstream.

shawkins:
@MikeEdgar @fvaleri @rareddy I was checking staging and production to confirm the current DisruptionsAllowed and it appears that the drain cleaner webhook is not seen as installed - that is the kafka resources do not have an explicit PDB with maxUnavailable set to 0.  However I do see the drain cleaner webhook configuration, so I'm not sure what is going on.  Can someone confirm we do indeed have an issue?

MikeEdgar:
My theory (untested) is that the order of add/delete events for the webhook during a bundle upgrade is such that the value of `drainCleanerWebhookFound` ends up as `false`.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
fvaleri: I see, but this is not how the node drain should work. I can understand that the autoscaler can't wait indefinitely, but it should at least be configurable.

The problem with this change is that the Drain Cleaner would still be notified and the cluster will still be rolled. This may not be a problem for a developer instance, but it's not desired for bigger instances.

shawkins:
> Perhaps when there is a change event we should list them and set the boolean if the result is non-empty.

If it's possible that there can be more than 1 of these at a time, then yes we could have a timing issue here - you'd have something like the event sequence add 1, add 2, delete 1, which would leave us in the false state.

> Perhaps when there is a change event we should list them and set the boolean if the result is non-empty.

Yes the informer cache can be consulted on delete.

I'll log a JIRA.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
shawkins: > I see, but this is not how the node drain should work.  I can understand that the autoscaler can't wait indefinitely, but it should at least be configurable.

There would need to be two changes here correct - one change to allow draining to even be attempted when disruptionsAllowed=0, and another to allow for the draining timeout to be configurable.

> The problem with this change is that the Drain Cleaner would still be notified and the cluster will still be rolled.

There are two cases here - staging clusters will have both developer and standard instances.  While production clusters should have only a single type.  Would you advocate in the longer term for production developer clusters to have the drain cleaner webhook not installed?

> but I guess it's not desired for bigger instances.

This change will only affect developer instances.  The standard plan will always have more than a single replica.

fvaleri: > There are two cases here - staging clusters will have both developer and standard instances. While production clusters should have only a single type. 

The OLM bundle should be smart enough to know if the drain cleaner have to be deployed or not based on the environment type. I don't think it has this capability, so we would need a new lightweight dev-prod bundle with no multi version and drain cleaner deployments.

> Would you advocate in the longer term for production developer clusters to have the drain cleaner webhook not installed?

Yes. The DC+CO draining will still work, but with single-broker instances there would be no benefit comparing to the standard K8s draining process.

rareddy:
@shawkins, @MikeEdgar  show we devise an alert when webhook is missing? we expect a drain cleaner always be available correct?

MikeEdgar:
> we expect a drain cleaner always be available correct?

At this point, yes. I'm not quite grasping the case where drain cleaner would not be present for developer instances (need to reread the thread above).

shawkins:
> show we devise an alert when webhook is missing? we expect a drain cleaner always be available correct?

Are there similar alerts for the other bundle components - operators, sync?

> At this point, yes. I'm not quite grasping the case where drain cleaner would not be present for developer instances (need to reread the thread above).

I'm gleaning that @fvaleri preferred solution longer-term in a production developer cluster is to not have the draincleaner installed at all, or at least the webhook uninstalled - as there's no value to having it running in that environment.

rareddy:
> Are there similar alerts for the other bundle components - operators, sync?

No, but with Sync, FSO we will see other stuff start failing, they are not enabled with an external trigger as drain cleaner is IMO.

>  as there's no value to having it running in that environment.

I am not sure about the amount of resource savings, but I would favor a consistent deployment model over small savings.

MikeEdgar:
> > as there's no value to having it running in that environment.
> 
> I am not sure about the amount of resource savings, but I would favor a consistent deployment model over small savings.

I agree, I would first want to look at proposing some selection criteria in drain cleaner to skip particular pods if necessary.

shawkins:
> No, but with Sync, FSO we will see other stuff start failing, they are not enabled with an external trigger as drain cleaner is IMO.

The validating webhook configuration is part of the bundle, just like the operator deployments.   So I'm just trying to clarify are you looking for some kind of alert that nothing from the bundle has been inadvertently deleted.

Or are you saying that you want a metric to be exposed from the FSO about whether it thinks the drain cleaner validating webhook configuration is installed, so that you can have an alert based upon that?

> I agree, I would first want to look at proposing some selection criteria in drain cleaner to skip particular pods if necessary.

We should separate out the longer-term issue.  I think we can move forward with this for now.

MikeEdgar:
> We should separate out the longer-term issue. I think we can move forward with this for now.

+1 - is this for 0.27.0?

shawkins:
> +1 - is this for 0.27.0?

I'd vote yes.

rareddy:
I thought webhook configuration is not part of the bundle but based on some environment specific config in OSD

> FSO about whether it thinks the drain cleaner validating webhook configuration is installed

I was thinking about this scenario, given above I am not sure it makes sense anymore. can you give the link to the bundle config for my understanding? thanks

shawkins:
> I was thinking about this scenario, given above I am not sure it makes sense anymore. can you give the link to the bundle config for my understanding? thanks

https://gitlab.cee.redhat.com/mk-ci-cd/rhosak-pipeline-configs/-/blob/rhosak-0.1-rhel-8/distgit/containers/rhosak-kas-strimzi-operator-bundle/src/templates/kas-strimzi-bundle.clusterserviceversion.yaml#L57

fvaleri:
+0 

Just for the record: my preference is not for saving some resources, but to avoid the race between K8s draining and drain cleaner, which may cause 2 rolling restarts (not a big issue for single-broker instances). If there is another way to avoid this without splitting the bundle (maybe a skip flag based on an env var), that would be the best.

rareddy:
@fvaleri with @shawkins modification this won't be enabled in a single broker environment avoiding the issue you mention. for standard sizes no change in current behavior. I am failing to see the concern, can you please explain

fvaleri:
@rareddy this change avoids the creation of the PDB required to block K8s draining when using the drain cleaner. Without the PDB, the pod eviction notifications will still reach the drain cleaner, which will annotate involved Kafka and ZK pods for restart.

shawkins:
https://issues.redhat.com/browse/MGDSTRM-9409 was created to address more completely.

PR #790, opened 2022-08-07
MGDSTRM-9388 making the logic tolerant to multiple webhook configs

shawkins:


shawkins:
Another thing to note here is that the addition or removal of the validating webhook configuration is not blast radius controlled - but that shouldn't be an issue as it's not expected to be dynamically removed.

MikeEdgar (approved):
LGTM, I confirmed that during an upgrade of a bundle there are indeed two `ValidatingWebhookConfiguration` and that with this PR the PDBs' `DisruptionAllowed` condition remains `False`.

Since it should almost never happen and would be helpful to know when it does, it would be worth logging at `INFO` when the `drainCleanerWebhookFound` is toggled.

shawkins:
> Since it should almost never happen and would be helpful to know when it does, it would be worth logging at INFO when the drainCleanerWebhookFound is toggled.

Updated to have info when turning it on, and warn when turning it off.

PR #794, opened 2022-08-11
remove now redundant broker plugins properties

k-wall:
The broker plugin was refactored to us avoid use of 'strimzi' in its configuration properties.  The version of the plugin that replied on the old name is no longer in use.  Thus the configuration using the old name can now be removed.


grdryn (approved):
LGTM, thanks @k-wall.

SamBarker (approved):
LGTM

PR #796, opened 2022-08-17
Set transaction.id.expiration.ms to 1 day

grdryn:
JIRA: https://issues.redhat.com/browse/MGDSTRM-9293

The default value for this is 7 days, but we want to reduce it to 1
day for the moment to reduce the chances that we'll hit an OOM as a
result of too many idempotent producer IDs being stored for too long.


MikeEdgar:
I think this might be better placed in the Strimzi version ConfigMap as an override so that it is set during the cluster upgrade. For example:

```yaml
  fleetshard_operands.yaml: |
    canary:
      init:
        image: null
      image: null
    admin-server:
      image: null
    kafka:
      brokerConfig:
        "transaction.id.expiration.ms": 86400000
```

PR #797, opened 2022-08-17
[MGDSTRM-8267] Set storage class by cluster provider type or default

MikeEdgar:


operator/src/main/java/org/bf2/operator/managers/InformerManager.java
```diff
@@ -104,6 +107,7 @@ protected void handleEvent(HasMetadata resource, ResourceAction action) {
 
         if (isOpenShift()) {
             routeInformer = resourceInformerFactory.create(Route.class, filterManagedByFleetshardOrStrimzi(openShiftSupport.adapt(kubernetesClient).routes()), eventSource);
+            infrastructureInformer = resourceInformerFactory.create(Infrastructure.class, openShiftSupport.adapt(kubernetesClient).config().infrastructures().withName("cluster"), eventSource);
```
shawkins, line 110: Unless there's a owner of the infrastructure, we don't want pass the eventSource here.

MikeEdgar:
> With these changes we probably don't need the storage class on the instance configuration any longer.
> 
> Alternatively it could be a little simpler to just set the storage class on the instance configuration when it's being loading and avoid any of the changes to how it's looked up. I don't think we'll need an informer on the Infrastructure in that scenario either.

That's a good point, I'll look at retrieving the storage class when initializing the instance configs rather than each time. I think we'd only want to do that when not already set.

operator/src/main/java/org/bf2/operator/managers/InformerManager.java
MikeEdgar: Removed the informer with the latest commit.

shawkins (approved):
LGTM

rareddy (approved):
lgtm

PR #798, opened 2022-08-24
MGDSTRM-9483 delaying deployment of admin/canary until kafka is ready

shawkins:
Given that node auto-scaling can introduce a delay in the deployment of the kafka, it would be best to wait until it is ready before creating the other deployments, so that any applicable timeouts start only at that point.  

The only difference to this approach vs now is there will be a small delay in the instance being seen as ready, and if the kafka ends up in an error state that will be reported as Ready=False Reason=Installing - because the admin / canary will still be in the installing state. Given the direction of MGDSTRM-9395 / MGDSTRM-9498 that is probably ok - the fleetmanager doesn't want to show the installation in an failed state regardless until the slo has expired.  However it hasn't been, and still won't be, clear the difference between installing and error - so we may eventually want to refine that with separate conditions.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/DeploymentOperand.java
+++ b/operator/src/main/java/org/bf2/operator/operands/DeploymentOperand.java
@@ -1,0 +1,58 @@
+package org.bf2.operator.operands;
+
+import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.apps.Deployment;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import org.bf2.common.OperandUtils;
+import org.bf2.operator.resources.v1alpha1.ManagedKafka;
+import org.bf2.operator.resources.v1alpha1.ManagedKafkaCondition.Status;
+
+import javax.inject.Inject;
+
+import java.util.Objects;
+
+public abstract class DeploymentOperand implements Operand<ManagedKafka> {
+
+    @Inject
+    protected KubernetesClient kubernetesClient;
+
+    @Inject
+    protected AbstractKafkaCluster kafkaCluster;
+
+    protected abstract Deployment cachedDeployment(ManagedKafka managedKafka);
+
+    protected void createOrUpdate(Deployment deployment) {
+        OperandUtils.createOrUpdate(kubernetesClient.apps().deployments(), deployment);
+    }
+
+    protected void createOrUpdate(Service service) {
+        OperandUtils.createOrUpdate(kubernetesClient.services(), service);
+    }
+
+    public abstract Deployment deploymentFrom(ManagedKafka managedKafka, Deployment current);
+
+    protected boolean handleReserveOrWaitForKafka(ManagedKafka managedKafka) {
+        Deployment current = cachedDeployment(managedKafka);
+
+        if (managedKafka.isReserveDeployment()) {
+            Deployment deployment = deploymentFrom(managedKafka, null);
+
+            deployment = ReservedDeploymentConverter.asReservedDeployment(current, deployment, managedKafka);
+            if (!Objects.equals(current, deployment)) {
+                createOrUpdate(deployment);
+            }
+            return true;
+        }
+
+        if (current == null) {
+            OperandReadiness kafkaReadiness = kafkaCluster.getReadiness(managedKafka);
+            if (kafkaReadiness.getStatus() != Status.True) {
+                // don't create until ready
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+}
--- a/operator/src/main/java/org/bf2/operator/operands/DeploymentOperand.java
+++ b/operator/src/main/java/org/bf2/operator/operands/DeploymentOperand.java
@@ -3,7 +3,6 @@ package org.bf2.operator.operands;
 import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.client.KubernetesClient;
-import io.strimzi.api.kafka.model.Kafka;
 import org.bf2.common.OperandUtils;
 import org.bf2.operator.resources.v1alpha1.ManagedKafka;
 
@@ -21,6 +20,8 @@ public abstract class DeploymentOperand implements Operand<ManagedKafka> {
 
     protected abstract Deployment cachedDeployment(ManagedKafka managedKafka);
 
+    public abstract Deployment deploymentFrom(ManagedKafka managedKafka, Deployment current);
+
     protected void createOrUpdate(Deployment deployment) {
         OperandUtils.createOrUpdate(kubernetesClient.apps().deployments(), deployment);
     }
@@ -28,8 +29,6 @@ public abstract class DeploymentOperand implements Operand<ManagedKafka> {
     protected void createOrUpdate(Service service) {
         OperandUtils.createOrUpdate(kubernetesClient.services(), service);
     }
-
-    public abstract Deployment deploymentFrom(ManagedKafka managedKafka, Deployment current);
 
     protected boolean handleReserveOrWaitForKafka(ManagedKafka managedKafka) {
         Deployment current = cachedDeployment(managedKafka);
@@ -44,15 +43,8 @@ public abstract class DeploymentOperand implements Operand<ManagedKafka> {
             return true;
         }
 
-        if (current == null) {
-            Kafka kafka = kafkaCluster.cachedKafka(managedKafka);
-            // if we don't yet have a clusterId, then we've never been ready
-            if (kafka == null || kafka.getStatus() == null || kafka.getStatus().getClusterId() == null) {
-                return true;
-            }
-        }
-
-        return false;
+        // if not yet created, wait until kafka is ready
+        return current == null && !kafkaCluster.hasKafkaBeenReady(managedKafka);
     }
 
 }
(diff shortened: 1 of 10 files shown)


shawkins:
A new corner case here is if the admin or the canary deployment is deleted when the kafka is not ready - they will not be added back until it is.  Since that is unlikely to occur, that may be ok.  The only way to workaround that is to discern if the kafka had ever been ready from the status - which I wasn't sure if it provides in an easily discernible way.

rareddy:
@shawkins how does the reconciliation happen after the first call, I mean after first create, kafka gets to ready state, how is update called so that admin/canary start deploying?

MikeEdgar:
> A new corner case here is if the admin or the canary deployment is deleted when the kafka is not ready - they will not be added back until it is. Since that is unlikely to occur, that may be ok.

Like you said this seems unlikely, fine with me. I think this change should also avoid creating the admin server route until the admin server is created.

> @machi1990 concern was with the canary init-container timeout. I think, but have not verified, that the canary init-container should not be impacted by a delay in bringing up the kafka.

This seems to be OK in my test of this change. There appears to even be the possibility that the canary-init finds the DNS record on the first attempt and skips any start-up delay. That could possibly be an issue - when it first runs it will wait 60s after finding the DNS record to ensure propagation to all authoritative nameservers.

shawkins:
Should be ready for consideration now.

rareddy (approved):
lgtm

MikeEdgar (approved):

shawkins:
> This seems to be OK in my test of this change. There appears to even be the possibility that the canary-init finds the DNS record on the first attempt and skips any start-up delay. That could possibly be an issue - when it first runs it will wait 60s after finding the DNS record to ensure propagation to all authoritative nameservers.

@grdryn @k-wall any additional thoughts on Mike's concern

k-wall:
I'm wondering whether the delete case is safe.  If the kafka instance is deleted before kafka has come up for the first time, org.bf2.operator.operands.KafkaInstance#delete will still try to delete the canary and admin server even though they have won't have been created.

I think `org.bf2.operator.operands.AbstractAdminServer#delete` is safe, as io.fabric8.kubernetes.client.dsl.Deletable#delete is safe to call on the resource that doesn't exist.

`org.bf2.operator.operands.AbstractCanary#delete` is safe, but the scale operation is going to fail.  There's a catch block in line 41 that will simply log the exception, but that will cause warn noise in the the logs.  

I also noticed (unrelated to this change), that the canary's service isn't explicitly deleted (the admin server cleans its up).

k-wall:
On the question of the DNS and this change, IIRC the idea of the [`DNS record found on first lookup` path ](https://gitlab.cee.redhat.com/mk-ci-cd/rhosak-pipeline-configs/-/blob/rhosak-0.1-rhel-8/distgit/containers/rhosak-strimzi-canary/canary-dns-init.sh#L49) was to optimise the case where the canary was restarted.  In that case, we reasoned that the canary will have previously gone through the `Retry domain lookup` path once, so we have the assurance that Route 53 propagation was (probably) complete.

With this change, as @MikeEdgar says above, it is probable that canary init will follow the `DNS record found on first lookup` path on first attempt as it is likely that zookeeper / kafka initialisation time will exceed the length of time for the registration in Route 53.  However, we lost the assurance of that 60s delay that was designed to allow propagation amongst the Route53 servers to occur.     We might see a regression with the return of the negative DNS caching problem within the cluster.   This will become more likely with developer instances, or as soon as we reach Kraft.

The "sleep 60" solution always seemed fragile.   I'd have always preferred the Rejected Alternative where Control Plane would signal to Data Plane once Route53 registration was complete.  https://architecture.appservices.tech/adr/76/#alternatives-considered-rejected

shawkins:
> There's a catch block in line 41 that will simply log the exception, but that will cause warn noise in the the logs.

This will occur currently today as a timing issue if there is additional reconciliation after the delete flag is first processed, so yes we should avoid that in general.

> I also noticed (unrelated to this change), that the canary's service isn't explicitly deleted (the admin server cleans its up).

It's probably best to make that change as well.  The namespace deletion is a catch all, but we've mostly followed the paradigm of whoever creates something is responsible for deleting.

> We might see a regression with the return of the negative DNS caching problem within the cluster. This will become more likely with developer instances, or as soon as we reach Kraft.

Based upon the comments of @katheris I don't think that is the case - the kafka resource will not add the listeners and clusterId until the kafka is ready.  So as this change currently stands we don't provide the routes to the control plane or start the canary deployment until that time - regardless of how long it takes for the kafka to become ready.  

This pr can then be further refined to instead look for the listeners (or clusterId) in the status, instead of the ready, and that will avoid the corner case.  That is with the blessing of the strimzi folks to make sure that seems like a stable solution.

Given that the canary-init nominally will start at the same time as the routes are reported to the control plane, is there still a concern about the first lookup succeeding and skipping the 60 second wait?  Or do we need to have a more precise mechanism, such as the response from the control plane?

k-wall:
> Based upon the comments of @katheris I don't think that is the case - the kafka resource will not add the listeners and clusterId until the kafka is ready. 

I'm not sure what listeners mean in this context. Strimzi creates the broker routes/services resources without waiting for the kafka and zookeeper is ready.  org.bf2.operator.controllers.ManagedKafkaController#updateManagedKafkaStatus will populate MK.status.routes immediatelythere are the expected number of routes are available.  I think this means Route53 registration will possibly start before the canary.  Whether this will actually be impactful or not, I'm not sure.

@grdryn WDYT?

shawkins:
> I'm not sure what listeners mean in this context. Strimzi creates the broker routes/services resources without waiting for the kafka and zookeeper is ready.

I mean that strimzi won't put them, or the clusterId, on the Kafka.status until the Kafka is ready.

shawkins:
> How is the Kafka CR relevant? I think updateManagedKafkaStatus is making observations of the OpenShift Route resources, and publishing MK.status.routes based on that information. Maybe I am missing something.

Sorry there are two paths here then.  The first path is route creation by strimzi and the reporting in the ManagedKafka status - that path is independent of the Kafka CR status.  I had mistakenly thought for a moment that we were waiting for the external listener to show up in the Kafka status for this, not just directly looking up the Routes.  

The other path is the addition of the listeners and clusterId to the Kafka status, that is only occurring after ready - 
https://github.com/strimzi/strimzi-kafka-operator/blob/64baa0984bc2225a54825182d76397a10228f76c/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java#L270 - which makes it a better check that the check for Ready than is currently in this pr.

Having these paths a separate does make it far more likely that the control plane will see the route information in the MK status well ahead of when the Kafka is ready and the canary / admin are created.

machi1990:
Just to add information on how the control plane handles the routes at the moment: 
It acts on the routes info only when it sees that the ManagedKafka status is `Ready=True`: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/a47e132fbbccb1b95a93b8d3e3888bc7faaf1a58/internal/kafka/internal/services/data_plane_kafka.go#L77-L83

This is when it'll attempt to create a route53 records and wait for them to be published before reporting that the kafka is `ready` to the end user.

machi1990 (approved):
lgtm

MikeEdgar:
> This is when it'll attempt to create a route53 records and wait for them to be published before reporting that the kafka is `ready` to the end user.

I think this removes any question on whether the Canary will find the DNS entry on the first try and skip the 60s wait (it won't). Thanks @machi1990

operator/src/main/java/org/bf2/operator/operands/DeploymentOperand.java
```diff
@@ -0,0 +1,58 @@
...
+
+    protected void createOrUpdate(Deployment deployment) {
+        OperandUtils.createOrUpdate(kubernetesClient.apps().deployments(), deployment);
+    }
+
+    protected void createOrUpdate(Service service) {
+        OperandUtils.createOrUpdate(kubernetesClient.services(), service);
+    }
+
+    public abstract Deployment deploymentFrom(ManagedKafka managedKafka, Deployment current);
```
grdryn, line 32: It might be slightly tidier to have this placed alongside the other declared abstract method on L22. WDYT?

grdryn (approved):
LGTM

shawkins:
> It acts on the routes info only when it sees that the ManagedKafka status is Ready=True

@machi1990 @MikeEdgar Based upon the concerns here, do we want to future proof this a little more and have the FSO only return the route information once kafka has been ready?

MikeEdgar:
> do we want to future proof this a little more and have the FSO only return the route information once kafka has been ready?

It makes sense to have it function that way. The cleanest would exclude the admin server route until that is ready also, but may not be worth the effort and it may interfere with the timing.

machi1990:
> > It acts on the routes info only when it sees that the ManagedKafka status is Ready=True
> 
> @machi1990 @MikeEdgar Based upon the concerns here, do we want to future proof this a little more and have the FSO only return the route information once kafka has been ready?

> It makes sense to have it function that way. The cleanest would exclude the admin server route until that is ready also, but may not be worth the effort and it may interfere with the timing.

+1 as well depending on how doable that is if it interferes with other timing issues as mentioned by Mike above probably something worth pursuing in a follow up JIRA?

k-wall (approved):

MikeEdgar (approved):
I haven't had a chance to try the latest commits yet, but everything logically looks good to me.

machi1990 (approved):

PR #801, opened 2022-08-29
MGDSTRM-9498-Bugs flagged from sonarcloud integration

biswassri:


operator/src/main/java/org/bf2/operator/events/ResourceEventSource.java
```diff
@@ -47,11 +47,10 @@ protected void handleEvent(HasMetadata resource, ResourceAction action) {
             if(resource.getMetadata().getOwnerReferences().isEmpty()) {
                 log.warnf("%s %s/%s does not have OwnerReference", resource.getKind(), resource.getMetadata().getNamespace(), resource.getMetadata().getName());
             } else {
-                getEventHandler().handleEvent(new ResourceEvent(action, ResourceID.fromFirstOwnerReference(resource).get()));
+                getEventHandler().handleEvent(new ResourceEvent(action, ResourceID.fromFirstOwnerReference(resource).orElse(null)));
```
MikeEdgar, line 50: Try something like this to handle the `Optional`:
```java
ResourceID.fromFirstOwnerReference(resource).ifPresentOrElse(
    ownerId -> getEventHandler().handleEvent(new ResourceEvent(action, ownerId)),
    () -> log.warnf("%s %s/%s does not have OwnerReference", resource.getKind(), resource.getMetadata().getNamespace(), resource.getMetadata().getName()));
```

operator/src/main/java/org/bf2/operator/operands/Canary.java
```diff
@@ -306,15 +306,20 @@ private List<EnvVar> buildEnvVar(ManagedKafka managedKafka, Deployment current)
         String kafkaVersion = managedKafka.getSpec().getVersions().getKafka();
         // takes the current Kafka version if the canary already exists. During Kafka upgrades it doesn't have to change, as any other clients.
         if (current != null) {
-            Optional<EnvVar> kafkaVersionEnvVar = current.getSpec().getTemplate().getSpec().getContainers().stream()
+            Optional<Container> kafkaVersionContainer = current.getSpec().getTemplate().getSpec().getContainers().stream()
```
MikeEdgar, line 309: Flat-mapping the stream to the container Envs avoids the intermediate `Optional` in the original code.
```java
Optional<EnvVar> kafkaVersionEnvVar = current.getSpec().getTemplate().getSpec().getContainers().stream()
        .filter(container -> "canary".equals(container.getName()))
        .map(Container::getEnv)
        .flatMap(Collection::stream)
        .filter(ev -> "KAFKA_VERSION".equals(ev.getName()))
        .findFirst();
if (kafkaVersionEnvVar.isPresent()) {
    kafkaVersion = kafkaVersionEnvVar.get().getValue();
}
```

perf/src/main/java/org/bf2/performance/framework/ManagedKafkaStateAssertionParameterResolver.java
```diff
@@ -69,7 +69,7 @@ public static Optional<Integer> getLargestKnownInstanceCount(TestInfo info) {
     }
 
     private static Stream<ManagedKafkaAssertion> getManagedKafkaAssertionStream(OpenshiftClusterData openshiftClusterData, Optional<Method> testMethod) {
-        return Arrays.stream(testMethod.get().getAnnotationsByType(ManagedKafkaAssertion.class))
+        return Arrays.stream(testMethod.orElse(null).getAnnotationsByType(ManagedKafkaAssertion.class))
```
MikeEdgar, line 72: Here you'll want to conditionally map the method to a stream of its `ManagedKafkaAssertion` annotations. Something like the following. Instead of `null`, the default will be an empty stream.
```java
return testMethod
    .map(m -> m.getAnnotationsByType(ManagedKafkaAssertion.class))
    .map(Arrays::stream)
    .orElseGet(Stream::empty)
    .filter(a -> a.instanceType().equals(openshiftClusterData.getWorkerInstanceType()))
    .sorted(Comparator.comparingInt(ManagedKafkaAssertion::minWorkers))
    .collect(Collectors.toList())
    .stream();
```

systemtest/src/main/java/org/bf2/systemtest/operator/FleetShardOperatorManager.java
```diff
@@ -56,9 +56,12 @@ public static CompletableFuture<Void> deployFleetShardOperator(KubeClient kubeCl
         }
         printVar();
         LOGGER.info("Installing {}", OPERATOR_NAME);
-
-        installedCrds =
-                Files.list(CRD_PATH).filter(p -> p.getFileName().toString().endsWith(CRD_FILE_SUFFIX)).collect(Collectors.toList());
+        Stream<Path> credFilePath = Files.list(CRD_PATH);
+        try {
```
MikeEdgar, line 60: This can be simplified further using try-with-resource by placing `Stream<Path> credFilePath = Files.list(CRD_PATH)` in the `try (...)` and removing the `finally` block.

MikeEdgar (changes requested):
Left a few comments. Generally, we won't be able to remove the `volatile` modifier without adding in synchronization.

operator/src/main/java/org/bf2/operator/events/ResourceEventSource.java
biswassri: @MikeEdgar This one is throwing an error of incorrect type. `Required type: ResourceID, Provided: void`

MikeEdgar: The code snippet should be the full contents within `if (getEventHandler() != null) { ... }`

biswassri: That worked!

operator/src/main/java/org/bf2/operator/operands/Canary.java
```diff
@@ -308,13 +309,16 @@ private List<EnvVar> buildEnvVar(ManagedKafka managedKafka, Deployment current)
...
+                    .map(Container::getEnv)
+                    .flatMap(Collection::stream)
                     .filter(ev -> "KAFKA_VERSION".equals(ev.getName()))
                     .findFirst();
-            if (kafkaVersionEnvVar.isPresent()) {
-                kafkaVersion = kafkaVersionEnvVar.get().getValue();
+
+                if (kafkaVersionEnvVar.isPresent()) {
+                    kafkaVersion = kafkaVersionEnvVar.get().getValue();
+
```
MikeEdgar, line 319: Formatting nit
```suggestion
            if (kafkaVersionEnvVar.isPresent()) {
                kafkaVersion = kafkaVersionEnvVar.get().getValue();
```

biswassri:
@shawkins Steve The `volatile` modifier was flagged as a bug by sonarcloud. So, just wanted to check if you're fine with removing it.

test/src/main/java/org/bf2/test/TestUtils.java
```diff
@@ -97,15 +97,15 @@ public void run() {
...
         return getLogPath(folderName, testClass, testMethod);
     }
 
-    public static Path getLogPath(String folderName, TestInfo info) {
+    public static Path getLogPath(String folderName, TestInfo info) throws NullPointerException {
         String testMethod = info.getDisplayName();
-        Class<?> testClass = info.getTestClass().get();
+        Class<?> testClass = info.getTestClass().orElse(null);
         return getLogPath(folderName, testClass, testMethod);
     }
```
MikeEdgar, line 110: The `orElse` paths here are guaranteed NPEs. Let's think about an alternate approach. Perhaps have the `getLogPath` called by these two methods to accept a `String` rather than the class name, then change these to map the class name orElse a constant "NOCLASS". For example:/
```java
String testClassName = info.getTestClass().map(Class::getName).orElse("NOCLASS");
```

perf/src/main/java/org/bf2/performance/InstanceProfiler.java
```diff
@@ -424,14 +424,13 @@ private void setup() throws Exception {
...
-    private AvailableResources getMinAvailableResources(Stream<Node> stream) {
+    private AvailableResources getMinAvailableResources(Stream<Node> stream) throws NullPointerException{
         AvailableResources resources = stream.map(TestUtils::getMaxAvailableResources)
                 .reduce((a1, a2) -> {
                     a1.cpuMillis = Math.min(a1.cpuMillis, a2.cpuMillis);
                     a1.memoryBytes = Math.min(a1.memoryBytes, a2.memoryBytes);
                     return a1;
-                })
-                .get();
+                }).orElse(null);
```
MikeEdgar, line 433: I don't think this is the right approach when the stream is empty. @shawkins , do you have any thoughts for a sensible identity value to pass to `reduce` to avoid the `Optional` here?

MikeEdgar (commented):
A few additional comments. The `volatile`s will need to stay. I'm leaning toward saying they're false positives.

test/src/main/java/org/bf2/test/TestUtils.java
biswassri: Right! But shouldn't we `log` it when it hits `NOCLASS`?

perf/src/main/java/org/bf2/performance/InstanceProfiler.java
```diff
@@ -562,7 +561,7 @@ protected LatencyResult determineLatency(Profile profile, Consumer<Workload> set
             result.medianEndToEndLatency99pct = TestUtils.getMedian(loadTestResult.endToEndLatency99pct);
             result.aggregatedPublishLatency50pct = loadTestResult.aggregatedPublishLatency50pct;
             result.aggregatedPublishLatency99pct = loadTestResult.aggregatedPublishLatency99pct;
-            result.maxConnectionCount = loadTestResult.additionalMetrics.get(KafkaBenchmarkDriverWithMetrics.CONNECTION_COUNT).stream().max(Double::compareTo).get();
+            result.maxConnectionCount = loadTestResult.additionalMetrics.get(KafkaBenchmarkDriverWithMetrics.CONNECTION_COUNT).stream().max(Double::compareTo).orElse(0.0);
```
biswassri, line 564: @MikeEdgar Also wanted to double-check this `orElse()` here. Does this seem fine?

perf/src/main/java/org/bf2/performance/InstanceProfiler.java
shawkins: @MikeEdgar I'm ok with just allowing the call to .get().  Forcing an alternative to the implicit NoSuchElementException will likely lead to adding a utility method that does exactly what you have as a replacement for get.

shawkins:
> Steve, the volatile modifier was flagged as a bug by sonar cloud. So, just wanted to check if you're fine with removing it.

@biswassri yes sonar has a bias against volatile.  Our usage is simply that we want the most up-to-date value and that the resource may be used by multiple threads.  In general it should not be removed unless synchronization is added, or it's converted to an AtomicReference.

perf/src/main/java/org/bf2/performance/InstanceProfiler.java
MikeEdgar: Yes, good point. @biswassri this can remain unchanged, but use `//NOSONAR` to silence the warning.

perf/src/main/java/org/bf2/performance/InstanceProfiler.java
MikeEdgar: @biswassri - sorry, forgot to comment here. Let's leave this as-is and add the `//NOSONAR` comment too.

biswassri: Sure!

MikeEdgar (approved):

kornys (approved):

PR #805, opened 2022-09-01
MGDSTRM-9275-Show Observabilty acceptence status in MKA

biswassri:


rareddy (approved):
lgtm

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaAgentController.java
```diff
@@ -112,6 +112,9 @@ private ManagedKafkaAgentStatus buildStatus(ManagedKafkaAgent resource) {
         } else {
             ConditionUtils.updateConditionStatus(readyCondition, statusValue, null, null);
         }
+        if (!this.observabilityManager.isObservabilityRunning() && readyCondition != null) {
```
MikeEdgar, line 115: The status will be non-null here. Removing the check should improve the coverage too :-)
```suggestion
        if (!this.observabilityManager.isObservabilityRunning()) {
```

MikeEdgar (approved):
LGTM

PR #806, opened 2022-09-02
MGDSTRM-9567 MGDSTRM-9555 allowing for the transition from 3 to 2

shawkins:
also simplifying the logic based upon the possibility of dynamic scaling and node balancing - since we are enforcing balanced node scaling and have the topology spread constraint / scaling factor on brokers, we shouldn't need any deep node logic here.

### DIFF
--- a/operator/src/test/java/org/bf2/operator/managers/IngressControllerManagerTest.java
+++ b/operator/src/test/java/org/bf2/operator/managers/IngressControllerManagerTest.java
@@ -125,7 +125,7 @@ public class IngressControllerManagerTest {
     }
 
     @Test
-    public void testReplicaReduction3to2() {
+    void testReplicaReduction3to2() {
         openShiftClient.resourceList((List)buildNodes(12)).createOrReplace();
 
         IntStream.range(0, 3).forEach(i -> {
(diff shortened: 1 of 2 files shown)


rareddy:
I guess this logic lets the replicas start at minimum of 1 node rather than minimum of 2.

shawkins:
> I guess this logic lets the replicas start at minimum of 1 node rather than minimum of 2.

No, we haven't gotten sign off on such a change yet.

There's a check that enforces the extra instance - https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/806/files#diff-89e55b846f9501475f8cd950173b4de15c56c06f2b9e53efee166862768000cdR502

operator/src/test/java/org/bf2/operator/managers/IngressControllerManagerTest.java
```diff
@@ -121,6 +124,32 @@ public void testReplicaReduction() {
...
+            mk.getSpec().getCapacity().setEgressPerSec(Quantity.parse("300Mi"));
+            Kafka kafka = this.kafkaCluster.kafkaFrom(mk, null);
+            openShiftClient.resource(kafka).createOrReplace();
+        });
+        informerManager.createKafkaInformer();
+
+        ingressControllerManager.reconcileIngressControllers();
+        checkAzReplicaCount(3);
+
+        // remove two kafkas - we should keep the same number of replicas
```
grdryn, line 145: Is this comment correct? My reading is that we expect 3 replicas on L143, but then 2 on L150, which I think contradicts the expectation of this comment?

shawkins: Sorry, that is a cut and paste leftover.  Will remove.

grdryn: Thanks!

grdryn (approved):
LGTM, thanks.

rareddy (approved):
lgtm

PR #809, opened 2022-09-08
Excluding getter-setter from sonar

biswassri:


api/src/main/java/org/bf2/operator/Generated.java
```diff
@@ -0,0 +1,15 @@
+package org.bf2.operator;
```
rareddy, line 1: IMO this class should probably reside in common module in some util or code package

MikeEdgar: We're already using Lombok for equals/hashcode implementations, I think I would rather use it for getters/setters than going with the annotation. The "generated" naming that is required for the annotation is a bit of a smell for code that isn't really generated.

biswassri: Oh, we are using Lombok? I couldn't find a Lombok.config so I thought this would be a better approach. @MikeEdgar  where should I look to find the lombok.config?

MikeEdgar: We're using it, but the config is optional unless you want to do something like used `lombok.addLombokGeneratedAnnotation = true`. You can add the config along with your change.

biswassri: Yeah I was hoping to add that, so for further classes, the annotation gets added. Does that sound good?

MikeEdgar: Yes, that's good. You can go ahead and remove any simple getter/setter methods in the `api` project and add the appropriate Lombok annotations to generate them as well.

api/src/main/java/org/bf2/operator/resources/v1alpha1/Versions.java
```diff
@@ -29,35 +31,14 @@ public class Versions {
     public static final String STRIMZI_CLUSTER_OPERATOR_V0_23_0_4 = "strimzi-cluster-operator.v0.23.0-4";
 
     @NotNull
+    @Getter @Setter
```
MikeEdgar, line 34: What happens with these on the class instead of being on the fields?

pom.xml
```diff
@@ -328,6 +328,7 @@
 	                    <!-- exclude entry points -->
 	                    <exclude>**/KasFleetShardOperator.class</exclude>
 	                    <exclude>**/Main.class</exclude>
+                        <exclude>**/**/ExcludedPOJO.class</exclude>
```
MikeEdgar, line 331: Is this needed?

api/src/main/java/org/bf2/operator/resources/v1alpha1/ManagedKafkaCapacity.java
```diff
@@ -18,14 +20,14 @@
...
-    private String maxDataRetentionPeriod;
-    private Integer maxConnectionAttemptsPerSec;
+    @Setter private Quantity ingressPerSec;
+    @Setter private Quantity egressPerSec;
+    @Getter @Setter private Quantity ingressEgressThroughputPerSec;
+    @Getter @Setter private Integer totalMaxConnections;
+    @Getter @Setter private Quantity maxDataRetentionSize;
+    @Getter @Setter private Integer maxPartitions;
+    @Getter @Setter private String maxDataRetentionPeriod;
+    @Getter @Setter private Integer maxConnectionAttemptsPerSec;
```
MikeEdgar, line 30: If these are on the class, I think Lombok should ignore the getter methods that already are implemented.

biswassri: Yeah that makes sense! Let me try that.

api/src/main/java/org/bf2/operator/resources/v1alpha1/Versions.java
biswassri: It generates for only those fields. But I think I can move it to class level here as well since the other variables are static and final.

biswassri:
@MikeEdgar I noticed, after I removed `ExcludedPOJO.class` from the pom the coverage went down by 8%, so I added it back. Also, trying to add `<sonar.issue.ignore.allfile>@Generated</sonar.issue.ignore.allfile>` for removing the generated class. I wanted to suggest, what do you think about removing `api` module completely as well from coverage?

MikeEdgar:
> after I removed `ExcludedPOJO.class` from the pom the coverage went down by 8%
There's something strange about that - there is no such class in the project. I suspect it is something else in that same commit that caused the decrease.

> what do you think about removing api module completely as well from coverage?
Once the getter/setter methods are excluded, there is very little code left and I think most of it is covered. One of the earlier pushes I think had the coverage up to 62% or so right?

biswassri:
> Once the getter/setter methods are excluded, there is very little code left and I think most of it is covered. One of the earlier pushes I think had the coverage up to 62% or so right?

Right! at commit `[81a3a6d](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/809/commits/81a3a6dbf05765d11eb3e4ffd311a2cc4234c6a7)` it went up to 62%. Once I moved all the `getter/setter` to a class level it went back to 59%. So I'm not sure, what's causing the various.

MikeEdgar (approved):
LGTM - coverage increases from 59% to almost 66%

rareddy (approved):
lgtm

PR #810, opened 2022-09-09
MGDSTRM-9508 making sure the developer instances are movable

shawkins:
PodSet pods won't have a parent deployment.  This annotation will allow them to still be movable by the auto-scaler.  This does not yet address a non-developer bin-packing scenario as we're not sure if that will be needed.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -475,6 +475,12 @@ public class KafkaCluster extends AbstractKafkaCluster {
 
         podTemplateBuilder.addAllToTolerations(OperandUtils.profileTolerations(managedKafka));
 
+        if (replicas == 1) {
+            podTemplateBuilder.editOrNewMetadata()
+                    .addToAnnotations("cluster-autoscaler.kubernetes.io/safe-to-evict", "true")
+                    .endMetadata();
+        }
+
         KafkaClusterTemplateBuilder templateBuilder = new KafkaClusterTemplateBuilder()
                 .withPod(podTemplateBuilder.build());
 
--- a/operator/src/test/resources/expected/developer-strimzi.yml
+++ b/operator/src/test/resources/expected/developer-strimzi.yml
@@ -212,6 +212,9 @@ spec:
           optional: false
     template:
       pod:
+        metadata:
+          annotations:
+            cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
         affinity:
           podAffinity:
             preferredDuringSchedulingIgnoredDuringExecution:


rareddy (commented):
small nit, maybe put the annotation string in the utility class that Mike created?

shawkins:
> small nit, maybe put the annotation string in the utility class that Mike created?

That seemed to be for our annotations, not the ones for third-party stuff - the strimzi, ingress, etc. annotations are not managed in that way yet.

MikeEdgar (approved):
LGTM

MikeEdgar:
> That seemed to be for our annotations, not the ones for third-party stuff - the strimzi, ingress, etc. annotations are not managed in that way yet.

That was the intent, yes. Specifically the `managedkafka.bf2.org` prefixed ones.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -475,6 +475,12 @@ private KafkaClusterTemplate buildKafkaTemplate(ManagedKafka managedKafka, int r
 
         podTemplateBuilder.addAllToTolerations(OperandUtils.profileTolerations(managedKafka));
 
+        if (replicas == 1) {
+            podTemplateBuilder.editOrNewMetadata()
+                    .addToAnnotations("cluster-autoscaler.kubernetes.io/safe-to-evict", "true")
```
machi1990, line 480: should the annotation be added to other pods as well of a developer instance:
- zookeeper 
- canary
- kafka-admin
- kafka-exporter
?

shawkins: The impetus here is the change to strimzi pod sets - the other pods will be managed by a deployment, which is fine for the auto-scaler to move.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -475,6 +475,12 @@ private KafkaClusterTemplate buildKafkaTemplate(ManagedKafka managedKafka, int r
 
         podTemplateBuilder.addAllToTolerations(OperandUtils.profileTolerations(managedKafka));
 
+        if (replicas == 1) {
```
machi1990, line 478: How about adding a `managedKafka.isDeveloperInstance()` method in `ManagedKafka` method that relies on https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/6aa4ac43d6eec97d2ba403bf8229269c2f5c85db/api/src/main/java/org/bf2/operator/resources/v1alpha1/ManagedKafka.java#L39?

shawkins: Not that we currently have additional profile plans, but the intent of checking the replicas instead of the profile directly was that we don't care about HA when the number of replicas is 1 and thus it's movable.  

As mentioned in the description we may eventually want to also tie this logic to whether bin-packing is being used or not.

rareddy (approved):
lgtm

PR #811, opened 2022-09-13
MGDSTRM-9666 removing labels that may change from reserved selectors

shawkins:
I was mistaken that this could be addressed by using a patch - the selectors are still immutable.  The proper way to do this is to just choose labels that won't change, similar to what is applied to our other deployments.


rareddy:
so if you have the right selectors then the PUT call works fine?

shawkins:
> so if you have the right selectors then the PUT call works fine?

Yes.  The problem here is having managedkafka.bf2.org/strimziVersion: "strimzi-cluster-operator.v0.23.0-4" in the selectors - I was just reusing the Kafka.metadata.  However you aren't allowed to change that after the deployment is created.

rareddy (approved):
lgtm

machi1990 (approved):

machi1990:
Thanks @shawkins @rareddy

MikeEdgar (approved):

machi1990:
Can this be included in the 0.29 release @rareddy @MikeEdgar @shawkins ?

PR #814, opened 2022-09-15
Fix upgrade test to use sync instead plain ManagedKafka CR

kornys:
This is fix for upgrade test. I see in jenkins that this test does not work because it used plain managedkafka CR instead of call to sync. With this change it works again.


MikeEdgar (approved):
LGTM :+1:

grdryn (approved):

PR #817, opened 2022-09-28
MGDSTRM-9590: add `num.recovery.threads.per.data.dir` config value

showuon:
What changes:
1. add `num.recovery.threads.per.data.dir` config value to broker
2. Calculate the `num.recovery.threads.per.data.dir` config value based on:
 CPU cores limits / storage volume count
3. We aggressively use the CPU limits, instead of CPU requested, because during log recovery, we should bring it back ASAP.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -123,7 +123,8 @@ public class KafkaCluster extends AbstractKafkaCluster {
     private static final String DO_NOT_SCHEDULE = "DoNotSchedule";
 
     private static final boolean DELETE_CLAIM = true;
-    // define jbod volume number to determin how many log.dirs we have
+    // define jbod volume number to determin how many log.dirs we have. So far, we only support 1 volume
+    // TODO: We should export it as a configurable parameter in the future
     private static final int JBOD_VOLUME_NUM = 1;
 
     private static final String KAFKA_EXPORTER_ENABLE_SARAMA_LOGGING = "enableSaramaLogging";
@@ -729,7 +730,7 @@ public class KafkaCluster extends AbstractKafkaCluster {
         BigDecimal cpuBytes = Quantity.getAmountInBytes(cpu);
         int cpuCores = Math.max(1, cpuBytes.intValue());
 
-        config.put("num.recovery.threads.per.data.dir", cpuCores);
+        config.put("num.recovery.threads.per.data.dir", cpuCores / JBOD_VOLUME_NUM);
 
         // Override broker config from operand override
         String strimzi = managedKafka.getSpec().getVersions().getStrimzi();
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -80,6 +80,7 @@ import javax.xml.bind.DatatypeConverter;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.math.BigDecimal;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -93,6 +94,8 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
+
+import static io.fabric8.kubernetes.api.model.Quantity.getAmountInBytes;
 
 /**
  * Provides same functionalities to get a Kafka resource from a ManagedKafka one
@@ -717,6 +720,19 @@ public class KafkaCluster extends AbstractKafkaCluster {
             config.put("cruise.control.metrics.topic.min.insync.replicas", this.configs.getConfig(managedKafka).cruiseControl.getMetricReporterTopicMinInsyncReplicas());
         }
 
+        Quantity cpu = new Quantity(kafkaConfigs.getContainerCpu());
+        BigDecimal cpuBytes = Quantity.getAmountInBytes(cpu);
+        log.error("!!! kafkaConfigs.getContainerCpu():" + kafkaConfigs.getContainerCpu());
+        log.error("!!! cpuBytes:" + cpuBytes);
+
+        int cpuCores = Math.max(1, cpuBytes.intValue());
+        log.error("!!! cpuCores:" + cpuCores);
+        System.out.println("!! kafkaConfigs.getContainerCpu():" + kafkaConfigs.getContainerCpu());
+        System.out.println("!! cpuBytes:" + cpuBytes);
+        System.out.println("!! cpuCores:" + cpuCores);
+
+        config.put("num.recovery.threads.per.data.dir", cpuCores);
+
         // Override broker config from operand override
         String strimzi = managedKafka.getSpec().getVersions().getStrimzi();
         Optional.ofNullable(this.overrideManager.getKafkaOverride(strimzi)).map(OperandOverrideManager.Kafka::getBrokerConfig).orElse(Map.of()).forEach((key, value) -> {
@@ -852,7 +868,7 @@ public class KafkaCluster extends AbstractKafkaCluster {
      * Get per broker value
      */
     private long getPerBrokerBytes(Quantity quantity, Supplier<String> defaultValue, int replicas) {
-        long bytes = Quantity.getAmountInBytes(Optional.ofNullable(quantity).orElseGet(() -> Quantity.parse(defaultValue.get()))).longValue();
+        long bytes = getAmountInBytes(Optional.ofNullable(quantity).orElseGet(() -> Quantity.parse(defaultValue.get()))).longValue();
         return bytes / replicas;
     }
 
@@ -871,7 +887,7 @@ public class KafkaCluster extends AbstractKafkaCluster {
      */
     private long calculateSafetyMargin(ManagedKafka managedKafka, Kafka current) {
         KafkaInstanceConfiguration config = this.configs.getConfig(managedKafka);
-        return Quantity.getAmountInBytes(config.getStorage().getMinMargin()).longValue()
+        return getAmountInBytes(config.getStorage().getMinMargin()).longValue()
                 + getIngressBytes(managedKafka, current) * config.getStorage().getCheckInterval()
                         * config.getStorage().getSafetyFactor();
     }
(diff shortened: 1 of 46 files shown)


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -815,19 +826,26 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current, long
...
 
-        PersistentClaimStorageBuilder builder = new PersistentClaimStorageBuilder()
-                .withId(JBOD_VOLUME_ID)
-                .withSize(getAdjustedMaxDataRetentionSize(current, storageBytes).getAmount())
-                .withDeleteClaim(DELETE_CLAIM);
+        List<SingleVolumeStorage> volumes = new ArrayList<>(volumeCount);
+
+        for (int jbodVolumeId = 0; jbodVolumeId < volumeCount; jbodVolumeId++) {
+            PersistentClaimStorageBuilder builder = new PersistentClaimStorageBuilder();
+            builder.withId(jbodVolumeId)
```
showuon, line 835: use the index as volume ID, i.e. 0, 1, 2, ...

operator/src/test/resources/expected/developer-strimzi.yml
```diff
@@ -115,6 +115,7 @@ spec:
       ssl.enabled.protocols: "TLSv1.3,TLSv1.2"
       client.quota.callback.static.storage.hard: "64424509440"
       leader.imbalance.per.broker.percentage: 0
+      num.recovery.threads.per.data.dir: 1
```
showuon, line 118: developer instance request 800m CPU, so the thread value should be 1.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -718,9 +720,18 @@ private Map<String, Object> buildKafkaConfig(ManagedKafka managedKafka, Kafka cu
...
         }
 
+        Quantity cpu = new Quantity(kafkaConfigs.getContainerCpu());
+        BigDecimal cpuBytes = Quantity.getAmountInBytes(cpu);
+        // cpuBytes might be less than 1
+        int cpuCores = Math.max(1, cpuBytes.intValue());
+
+        int volumeCount = kafkaConfigs.getVolumeCount();
+        // since the thread number is per data dir, we should consider the volume count in the broker
+        config.put("num.recovery.threads.per.data.dir", cpuCores / volumeCount);
```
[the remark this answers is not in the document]
grdryn: Looking at the [Kafka docs](https://kafka.apache.org/documentation/#brokerconfigs_num.recovery.threads.per.data.dir), this must indeed be an int, but it looks like only positive ints are valid `[1,...]`. I think there are cases where this calculcation could result with `0`, which sounds like an invalid value on the broker.

E.g. if we have 4 CPU cores, but 5 volumes, then the result here will be `0` (or any time we have more volumes than CPU cores. Is there anything preventing us from such a scenario in the future? I think we should maybe say that this needs to be at least 1. WDYT?

I wonder if it would make sense to not force `cpuCores` to be an int, and not force it to be 1, then allow decimal division here, and then only round the result in the end (and make sure it's at least 1)?

grdryn (commented):
I left a small comment about a slight worry around the calculation. I'll also add @shawkins and @MikeEdgar as reviewers from the fleetshard side :)

MikeEdgar (commented):
Any concerns for the blast radius when deploying the operator?

robobario:
Why introduce configurable volume count in this PR? I can see we would need to remember to modify this method in future if we wanted to support multiple volumes, but for now it's made the calculation a little more complex than it would be with the existing single-volume support. Is there a plan to support multiple volumes soon?

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
```diff
@@ -276,6 +277,25 @@ void testManagedKafkaToKafkaWithSizeChanges() throws IOException {
...
+        //When
+        Kafka kafka = kafkaCluster.kafkaFrom(managedKafka, null);
+
+        //Then
+        final String actualRecoveryThreadNum = String.valueOf(kafka.getSpec().getKafka().getConfig().get("num.recovery.threads.per.data.dir"));
+
+        Quantity cpu = new Quantity(configs.getConfig(managedKafka).getKafka().getContainerCpu());
+        BigDecimal cpuBytes = Quantity.getAmountInBytes(cpu);
+        int cpuCores = Math.max(1, cpuBytes.intValue());
+        assertEquals(String.valueOf(cpuCores/volumeCount), actualRecoveryThreadNum);
```
SamBarker, line 296: This effectively reproduces the production code so to me it's not a valid test.

Personally I'd hard code the test expectations to ensure the production code generates the correct results.

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
```diff
@@ -276,6 +277,25 @@ void testManagedKafkaToKafkaWithSizeChanges() throws IOException {
         diffToExpected(larger, "/expected/strimzi.yml", "[{\"op\":\"replace\",\"path\":\"/spec/kafka/config/client.quota.callback.static.storage.soft\",\"value\":\"28633115306\"},{\"op\":\"replace\",\"path\":\"/spec/kafka/config/client.quota.callback.static.storage.hard\",\"value\":\"28633115306\"},{\"op\":\"replace\",\"path\":\"/spec/kafka/storage/volumes/0/size\",\"value\":\"41383100373\"}]");
     }
 
+    @Test
+    void numRecoveryThreadShouldConsiderStorageCount() throws IOException {
```
SamBarker, line 281: I think this misses the basic test case something along the lines of 

```java
    @Test
    void shouldUseOneThreadPerCore() throws IOException {
        //Given
        final var coreCount = new Quantity("8000m");
        alternativeConfig(clone -> clone.getKafka().setContainerCpu(coreCount.toString()));
        final ManagedKafka managedKafka = dummyManagedKafka("0");

        //When
        Kafka kafka = kafkaCluster.kafkaFrom(managedKafka, null);

        //Then
        final String actualRecoveryThreadNum = String.valueOf(kafka.getSpec().getKafka().getConfig().get("num.recovery.threads.per.data.dir"));

        assertEquals(String.valueOf(Quantity.getAmountInBytes(coreCount).intValue()), actualRecoveryThreadNum);
    }
```

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
SamBarker, line 733: I've always subscribed to the rule of thumb when doing IO heavy workloads (mostly maven builds to be fair) of using two threads per core.  So I'm tempted to say we should set `Math.floor(2 * (cpuCores / volumeCount))`.

The other thought that goes through my head here, is this formula assumes each volume is equally important & equally large (needing an equal share of the resources). When ever I've considered multiple volume support in RHOSAK I've been thinking about a customer data Dir and a service data dir. I appreciate that giving 80% of the cores to the customer data dir and 20% to the service is not an option given the Kafka API I wonder if its better to just set the value to core count and let the additional threads for the service data dir idle out?

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -815,19 +826,26 @@ private Storage buildKafkaStorage(ManagedKafka managedKafka, Kafka current, long
...
-        PersistentClaimStorageBuilder builder = new PersistentClaimStorageBuilder()
-                .withId(JBOD_VOLUME_ID)
-                .withSize(getAdjustedMaxDataRetentionSize(current, storageBytes).getAmount())
-                .withDeleteClaim(DELETE_CLAIM);
+        List<SingleVolumeStorage> volumes = new ArrayList<>(volumeCount);
+
+        for (int jbodVolumeId = 0; jbodVolumeId < volumeCount; jbodVolumeId++) {
+            PersistentClaimStorageBuilder builder = new PersistentClaimStorageBuilder();
+            builder.withId(jbodVolumeId)
+                    .withSize(getAdjustedMaxDataRetentionSize(current, storageBytes, jbodVolumeId).getAmount())
```
SamBarker, line 836: This assumes we want each volume to have an equal size how valid is that assumption?

operator/src/main/java/org/bf2/operator/operands/KafkaInstanceConfiguration.java
```diff
@@ -177,6 +177,9 @@ public static class Kafka extends Container {
         protected String egressPerSec;
         @JsonProperty("storage-class")
         protected String storageClass;
+        // jbod volume count is to determine how many log.dirs we have. Default is 1
+        @JsonProperty("volume-count")
+        protected int volumeCount = 1;
```
SamBarker, line 182: For multi volume support I think we really want something more flexible such as 

```json
{
  "volumes": [
    {
      "id": "customerDataDir",
      "capacity": "333gi"
    },
    {
      "id": "serviceDataDir",
      "capacity": "50gi"
    }
  ]
}
```

SamBarker (commented):
I'm not sure that the multiple volume support is flexible enough, its not clear to me why we are introducing the concept of additional volumes to RHOASK as part of this PR. I think that needs additional thought.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
k-wall: I don't think we want to expand the scope of this PR to allow the number of volumes to be changed.  That's would be a significant change which would need to be prioritised separately.

With my earlier comment re. factoring in number of storage volumes into the calculation, I just meant just use `JbodStorage#getVolumes().size()` as a factor in the calculation, or perhaps just make it so the function that calculates the number of threads fail if ever that invariant changes:

```
if (JbodStorage#getVolumes().size() > 1) {
  throw new IllegalStateException ("This algorithm doesn't support support > 1 volumes");
} 
```

showuon: Ah, I see, sorry for misunderstanding!

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -900,10 +911,11 @@ private Quantity getAdjustedMaxDataRetentionSize(Kafka current, long storageByte
             Storage storage = current.getSpec().getKafka().getStorage();
             if (storage instanceof JbodStorage) {
                 JbodStorage jbodStorage = (JbodStorage)storage;
+                List<SingleVolumeStorage> volumes = jbodStorage.getVolumes();
```
grdryn:
```suggestion
```

It looks like this variable is not used.

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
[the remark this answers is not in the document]
showuon: Good point! I've updated it with:
1. use `double` for `cpuCores`
2. use Math.max(1, cpuCores/volumes) to avoid 0 result.

 Thanks.

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
showuon: Good point! Thanks for the suggestion!

operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
showuon: Thanks for the suggestion. I'll stick to 1 thread per core, unless we agree that's what we expected. And also revert the "partial" multiple volume support.

showuon:
> Any concerns for the blast radius when deploying the operator?

No, that will not happen since this config only take effect when unclean shutdown. 

And also revert the multiple volumes support in the latest commit. Thanks.

grdryn (approved):
LGTM!

k-wall (approved):

PR #818, opened 2022-09-28
refactorings to support testing with a kas-installer instance

shawkins:
also creating a more balanced test for partitions so that there's additional consumer groups

### DIFF
--- a/perf/src/main/java/org/bf2/performance/InstanceProfiler.java
+++ b/perf/src/main/java/org/bf2/performance/InstanceProfiler.java
@@ -112,20 +112,20 @@ public class InstanceProfiler {
 
     // recommended latency optimized profile
     public static final Profile LATENCY = new Profile("acks=1\n", "auto.offset.reset=earliest\nenable.auto.commit=false\n",
-                "compression.type=uncompressed\ncleanup.policy=delete\nretention.ms=240000\n");
+                "cleanup.policy=delete\nretention.ms=240000\n");
     // recommended latency optimized profile, but with batching disabled for throughput testing
     public static final Profile LATENCY_NO_BATCHING = new Profile("acks=1\nbatch.size=0\n", "auto.offset.reset=earliest\nenable.auto.commit=false\n",
-                "compression.type=uncompressed\ncleanup.policy=delete\nretention.ms=240000\n");
+                "cleanup.policy=delete\nretention.ms=240000\n");
 
     // using both https://www.confluent.io/blog/configure-kafka-to-minimize-latency/
     // and https://docs.confluent.io/cloud/current/client-apps/optimizing/throughput.html#optimizing-for-throughput
     public static final Profile THROUGHPUT = new Profile("acks=1\nbatch.size=200000\nlinger.ms=100\n",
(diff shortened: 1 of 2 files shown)


perf/src/main/java/org/bf2/performance/InstanceProfiler.java
```diff
@@ -311,8 +317,8 @@ public static void main(String[] args) throws Exception {
     private void runLocalTest() throws Exception {
         Profile profile = LATENCY;
         sizeAndUpdateConfig();
-        deployIfNeeded("profile");
-        OMBDriver driver = createDriver(instanceBootstrap, profile);
+        deployIfNeeded();
+        updateDriver(profile);
 
         byte[] storeBytes = Base64.getDecoder().decode(kafkaProvisioner.getTlsConfig().getTrustStoreBase64());
         File store = new File("target", "listener.jks");
```
MikeEdgar, line 324: It doesn't need to be fixed here, but when I used this method locally, I needed to replace the keystore load with first reading `ssl.truststore.location` from the `appServicesProperties`.

MikeEdgar (commented):
Should we also remove the `compression.type=uncompressed` from the topic configs? It will be a certain problem when using newer versions of the topic policy plugin. Easy enough to patch it if necessary though.

I can take a bit of time tomorrow and spin up some clusters to try it.

perf/src/main/java/org/bf2/performance/InstanceProfiler.java
shawkins: Yes, I haven't ever run the local test execution against a kas-installer cluster.  This method is mostly for debugging purposes, so I'm fine with whatever changes here.

shawkins:
> Should we also remove the compression.type=uncompressed from the topic configs?

That's fine too.  I had briefly been using the compression scenario to test the affects of storage bandwidth - compressed from the producer and uncompressed at the broker leading to very high EBS demand, but of course that will no longer be allowed.  OMB now has support for randomization for testing the compressed scenarios, which I haven't tried out yet.

shawkins:
Updated with the explicit topic compression removed.

MikeEdgar (approved):
Haven't had a chance to try it myself yet, but all looks good. Something I'd like to look at in the future is skipping all the kafka-client setup when the intent is to use a local worker.

rareddy (approved):
lgtm

perf/src/main/java/org/bf2/performance/ManagedKafkaProvisioner.java
```diff
@@ -520,18 +517,23 @@ ManagedKafkaDeployment deployCluster(String namespace, ManagedKafka managedKafka
...
+        Resource<ManagedKafka> mkResource = cluster.kubeClient()
+                .client()
+                .resources(ManagedKafka.class)
+                .inNamespace(KAFKA_NAMESPACE)
+                .withName(name);
+        try {
+            return mkResource.get();
+        } catch (KubernetesClientException e) {
+
+        }
```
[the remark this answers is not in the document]
shawkins: I'm not sure why that was added, removed now.

PR #820, opened 2022-09-30
MGDSTRM-9815 accounting for local storage / emptyDir wrt autoscaling

shawkins:


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -618,7 +624,14 @@ private KafkaExporterSpec buildKafkaExporter(ManagedKafka managedKafka) {
...
                 .withImage(this.overrideManager.getKafkaExporterImage(strimzi).orElse(null))
-                .withResources(config.getExporter().buildResources());
+                .withResources(config.getExporter().buildResources())
+                .withNewTemplate()
+                .withNewPod()
+                .withNewMetadata()
+                .withAnnotations(SAFE_TO_EVICT)
+                .endMetadata()
+                .endPod()
+                .endTemplate();
```
MikeEdgar, line 634: Just confirming - Exporter and Cruise Control are backed by `Deployment`s, so this is due to local storage?
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

shawkins: Yes, see the pods listed under https://issues.redhat.com/browse/MGDSTRM-9741

rareddy:
just to make sure Admin, Canary and CruiseControl do not use "emptydir" so they do not need the annotation? or should we set one for consistency?

MikeEdgar (approved):

shawkins:
> just to make sure Admin, Canary and CruiseControl do not use "emptydir" so they do not need the annotation? 

Admin and canary do not.  CruiseControl does, it's part of these changes.

> or should we set one for consistency?

I assumed that would be done with any later changes that introduced emptyDirs, but we can certainly add it proactively if you want.  The only downside is that it will bounce those pods - but that shouldn't be a concern with admin / canary.

rareddy:
> CruiseControl does, it's part of these changes.

Sorry I missed seeing that.

> I assumed that would be done with any later changes that introduced emptyDirs,

agree

rareddy (approved):
lgtm

PR #821, opened 2022-09-30
initial commit to force the ingress controller count to be 1 

rareddy:
initial commit to force the ingress controller count to be 1  to reduce the costs

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -570,7 +570,7 @@ public class IngressControllerManager {
     int numReplicasForZone(LongSummaryStatistics ingress, LongSummaryStatistics egress,
             long connectionDemand, double zonePercentage) {
         // use the override if present
-        int desiredReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
+        int minimumReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
         if (azReplicaCount.isPresent()) {
             return azReplicaCount.get();
         }
@@ -600,7 +600,7 @@ public class IngressControllerManager {
         int replicaCount = (int)Math.ceil(throughputDemanded / throughputPerIngressReplica);
         int connectionReplicaCount = numReplicasForConnectionDemand((long) (connectionDemand * zonePercentage));
 
-        return Math.max(desiredReplicaCount, Math.max(connectionReplicaCount, replicaCount));
+        return Math.max(minimumReplicaCount, Math.max(connectionReplicaCount, replicaCount));
     }
 
     static LongSummaryStatistics summarize(List<Kafka> kafkas, Function<Kafka, String> quantity,
@@ -617,20 +617,20 @@ public class IngressControllerManager {
 
     int numReplicasForDefault(long connectionDemand) {
         // use the override if present
-        int desiredReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
+        int minimumReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
         if (defaultReplicaCount.isPresent()) {
             return defaultReplicaCount.get();
         } else if (nodeInformer.getList().size() > 3){
             // enforce a minimum of two replicas on clusters that can accommodate it when no default specified
             // which may change if we don't want to
             // provide pod / node level HA for the az specific replicas.
-            desiredReplicaCount = 2;
+            minimumReplicaCount = 2;
         }
 
         /*
          * an assumption here is that these ingress replicas will not become bandwidth constrained - but that may need further qualification
          */
-        return Math.max(desiredReplicaCount, numReplicasForConnectionDemand(connectionDemand));
+        return Math.max(minimumReplicaCount, numReplicasForConnectionDemand(connectionDemand));
     }
 
     private int numReplicasForConnectionDemand(double connectionDemand) {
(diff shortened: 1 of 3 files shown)


operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -576,11 +562,22 @@ private void buildIngressController(String name, String domain,
...
+        azReplicaCount = value;
+    }
+
     int numReplicasForZone(LongSummaryStatistics ingress, LongSummaryStatistics egress,
             long connectionDemand, double zonePercentage) {
         // use the override if present
+        int desiredReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
         if (azReplicaCount.isPresent()) {
             return azReplicaCount.get();
+        } else if (nodeInformer.getList().size() > 3){
```
shawkins, line 576: This should be removed - we don't want to force two replicas anymore for each az.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -492,18 +490,6 @@ private void buildIngressController(String name, String domain,
 
         Optional<IngressController> optionalExisting = Optional.ofNullable(existing);
         IngressControllerBuilder builder = optionalExisting.map(IngressControllerBuilder::new).orElseGet(IngressControllerBuilder::new);
-        Integer existingReplicas = optionalExisting.map(IngressController::getSpec).map(IngressControllerSpec::getReplicas).orElse(null);
```
shawkins, line 495: Note to others - we decided that this can be removed for now as we'll likely run with a single replica most of the time.  Until we are dealing with larger instance counts, or we need to support a higher amount of peak demand - we'll run with 1.  So it's highly unlikely that we need to dampen the scale down.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
rareddy: removed

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -625,14 +617,15 @@ static LongSummaryStatistics summarize(List<Kafka> kafkas, Function<Kafka, Strin
...
+        int desiredReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
         if (defaultReplicaCount.isPresent()) {
             return defaultReplicaCount.get();
         }
 
         /*
          * an assumption here is that these ingress replicas will not become bandwidth constrained - but that may need further qualification
          */
-        return numReplicasForConnectionDemand(connectionDemand);
+        return Math.max(desiredReplicaCount, numReplicasForConnectionDemand(connectionDemand));
```
shawkins, line 628: @rareddy but you want to still bump it up to 2 here, what you previously had should work.

rareddy: oops, misunderstood the previous comment, restored it now, and added a test for it.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -576,9 +562,15 @@ private void buildIngressController(String name, String domain,
...
 
+    // for testing
+    void setAzReplicaCount(Optional<Integer> value) {
+        azReplicaCount = value;
+    }
+
     int numReplicasForZone(LongSummaryStatistics ingress, LongSummaryStatistics egress,
             long connectionDemand, double zonePercentage) {
         // use the override if present
+        int desiredReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
```
MikeEdgar, line 573: Could this be called `minimumReplicaCount` ?

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -625,14 +617,20 @@ static LongSummaryStatistics summarize(List<Kafka> kafkas, Function<Kafka, Strin
 
     int numReplicasForDefault(long connectionDemand) {
         // use the override if present
+        int desiredReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
```
MikeEdgar, line 620: Same as above re: name

MikeEdgar (approved):
lgtm, just a couple comments on the var name for clarity

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
rareddy: corrected.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
rareddy: corrected.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -625,14 +617,20 @@ static LongSummaryStatistics summarize(List<Kafka> kafkas, Function<Kafka, Strin
 
     int numReplicasForDefault(long connectionDemand) {
         // use the override if present
+        int minimumReplicaCount = nodeInformer.getList().size() > 0 ? 1:0;
         if (defaultReplicaCount.isPresent()) {
             return defaultReplicaCount.get();
+        } else if (nodeInformer.getList().size() > 3){
+            // enforce a minimum of two replicas on clusters that can accommodate it when no default specified
+            // which may change if we don't want to
+            // provide pod / node level HA for the az specific replicas.
```
shawkins, line 626: This method is specifically for the bootstrap / default, which we still want to be HA.

shawkins (approved):
LGTM

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
shawkins, line 626: This method is specifically for the bootstrap / default, which we still want to be HA.

```suggestion
```

PR #822, opened 2022-10-03
MGDSTRM-9799 adding overrides for standard / r5.xlarge

shawkins:
We don't have a great mechanism for this - it's dependent upon both profile type and strimzi version.  So for now I'm just creating a one-off that we can refine later.

I've also refined the initially proposed values slightly so that the cpu request of the broker will fit on an r5.xlarge that has a max allocatable cpu of 3.5.  The cpu limit was then raised separately, so the numbers will still be inline with what was reported.


rareddy (approved):
lgtm

PR #825, opened 2022-10-12
Increase connection.failed.authentication.delay.ms to 1000

robobario:
Why:
If a client uses incorrect credentials, or there is some other problem with authentication this will reduce pressure on any backing authentication services.


showuon (approved):

rareddy (approved):
lgtm

rareddy (approved):
lgtm

PR #826, opened 2022-10-13
[WIP] Fix: Report aggregation for jacoco

biswassri:


pom.xml
```diff
@@ -331,12 +331,19 @@
...
+                        <goals>
+                            <goal>report-aggregate</goal>
+                        </goals>
+                        <configuration>
+                            <includeCurrentProject>true</includeCurrentProject>
+                            <formats>
+                                <format>XML</format>
+                            </formats>
+                        </configuration>
+                    </execution>
```
MikeEdgar, line 346: The `sonar.yml` job is looking for the `jacoco.xml` files in `target/site/jacoco` and `report-aggregate` puts them in `target/site/jacoco-aggregate`. You should be able to add `<outputDirectory>${project.reporting.outputDirectory}/jacoco</outputDirectory>` to override it to the expected location.

Also, just a nit - but this XML block can be indented.

MikeEdgar (changes requested):

pom.xml
biswassri: Oh yes! I missed that change in this commit. But @MikeEdgar one concern/question I have is that 
Even with these changes, in the operator `index.html` we see `api` and `common` coverage but not `operator` coverage. Same for `sync`. I think we are missing something here. What do you think?

I experimented with `report` and `report-aggregate` but had no luck.

MikeEdgar: Something is definitely still not right... I'm checking now too.

MikeEdgar:
FYI, version 0.8.9 of the JaCoCo plugin was released.

MikeEdgar:
@biswassri , that repo is a not current. See [Maven Central](https://search.maven.org/artifact/org.jacoco/jacoco-maven-plugin). It was only just tagged a few days ago.

biswassri:
> @biswassri , that repo is a not current. See [Maven Central](https://search.maven.org/artifact/org.jacoco/jacoco-maven-plugin). It was only just tagged a few days ago.

Oh okay! Thanks for letting me know.

PR #828, opened 2022-10-27
MGDSTRM-9975 ingress controller updates

shawkins:
- removing cpu override for the kas/default ic
- passing the maxConnections from the config into the ingresscontroller
- generalizing how we set tuningOptions/unsupportedProperties to be forward compatible

I'm not sure what a good initial value for maxConnections should be - I've set it so that we could handle 36 SU at 1500 connections per broker with just a single replica.  The peak percentage will then allow for a demand of up 75000 connections per replica (the number we were assuming before) - both of those values can change of course pending feedback.


shawkins:
Unfortunately these test errors indicate we have some test flakiness - https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/actions/runs/3346120820/jobs/5542526753#step:5:4809

With fabric8 6+ we'll be able to better isolate tests as we can easily reset the state of the mock server.

grdryn (approved):

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -285,7 +289,12 @@ public void onDelete(IngressController obj, boolean deletedFinalStateUnknown) {
         // this is to patch the IngressController Router deployments to correctly size for resources, should be removed
         // after https://issues.redhat.com/browse/RFE-1475 is resolved.
         if (deploymentResourceBuilder.hasLimits() || deploymentResourceBuilder.hasRequests()) {
-            this.deploymentResourceRequirements = deploymentResourceBuilder.build();
+            this.azDeploymentResourceRequirements = deploymentResourceBuilder.build();
+
+            // use the default cpu for the kas/default ic
+            deploymentResourceBuilder.addToRequests(CPU, Quantity.parse("100m"));
```
rareddy, line 295: so this is default CPU you were mentioning in the comment?

shawkins: Yes, we're putting it in here explicitly so that it overwrites the 1700m we currently have set.  In a later release, we can remove this.

shawkins:
Tested against my local cluster and everything looks good.  If there are no objections, I'll merge this.

PR #829, opened 2022-11-03
MGDSTRM-10039 elevating the priority of broker and zookeeper pods

shawkins:
also adding some trivial instance profiler changes.

### DIFF
--- a/operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
+++ b/operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
@@ -217,6 +217,18 @@ class KafkaClusterTest {
     }
 
     @Test
+    void testElevatedPriority() throws IOException {
+        final ManagedKafka managedKafka = exampleManagedKafka("2Ti");
+
+        Mockito.when(overrideManager.useElevatedPriority(Mockito.anyString())).thenReturn(true);
+
+        Kafka kafka = kafkaCluster.kafkaFrom(managedKafka, null);
+
+        assertEquals(KafkaCluster.KAS_FLEETSHARD_MEDIUM_PRIORITY, kafka.getSpec().getKafka().getTemplate().getPod().getPriorityClassName());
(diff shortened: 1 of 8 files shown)


shawkins:
I'm not sure what a good naming convention is here.  We already have kas-fleetshard-high.  I went with core, but you could also go with standard, moderate, medium high, etc.

The choice of 10 for the priority value is also arbitrary.  We can of course space that out more if desired - high is 1000000

MikeEdgar (approved):
lgtm. I'm not sure about the naming either. I don't object to using `core`. Maybe something like `operand-high` or similar would be more descriptive?

rareddy:
Since "high" name already set precedent, I like the @shawkins idea of going with "standard, moderate, medium high, etc." then there is no ambiguity when referred in conjunction with other class types.

operator/src/test/java/org/bf2/operator/operands/KafkaClusterTest.java
```diff
@@ -216,6 +216,18 @@ private void alternativeConfig(Consumer<KafkaInstanceConfiguration> configModifi
...
 
+    @Test
+    void testElevatedPriority() throws IOException {
+        final ManagedKafka managedKafka = exampleManagedKafka("2Ti");
+
+        Mockito.when(overrideManager.useElevatedPriority(Mockito.anyString())).thenReturn(true);
+
+        Kafka kafka = kafkaCluster.kafkaFrom(managedKafka, null);
+
+        assertEquals("kas-fleetshard-core", kafka.getSpec().getKafka().getTemplate().getPod().getPriorityClassName());
```
rareddy, line 227: I guess these need to be updated to new name to fix the unit tests

shawkins: Yes, sorry things were incomplete.  Should be good now.

rareddy (approved):
lgtm

PR #833, opened 2022-11-13
MGDSTRM-10069 updating the properties based upon additional testing

shawkins:
Based upon some additional scale testing I'm seeing no worse than 100MiB per 10000 connections, so we should be good to lower our memory request to 1 gig instead of 2.  We should also try to satisfy the connection demand for ~36 SU to match our current assumptions about bandwidth at a rate of 100% peak (locally I actually had to set the max connections well above the desired level to not see any errors - so even the 100% rate could be conservative).

### DIFF
--- a/operator/src/main/resources/application.properties
+++ b/operator/src/main/resources/application.properties
@@ -27,7 +27,7 @@ mock.factory.interval=15s
 
 #ingress controller resources - an alternative profile can create fewer/smaller
 ingresscontroller.request-cpu=1700m
-ingresscontroller.request-memory=2Gi
+ingresscontroller.request-memory=1Gi
 #ingresscontroller.default-replica-count=1
 #ingresscontroller.az-replica-count=0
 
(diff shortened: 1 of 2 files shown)


MikeEdgar (approved):
lgtm

rareddy:
is my assumption true that with this we never probably have to scale AZ specific Ingress Controllers beyond 1 per AZ when limited to 36 SU if we limit the connections to 9000/SU. I see 9000*36/3 = 10800

shawkins:
> is my assumption true that with this we never probably have to scale AZ specific Ingress Controllers beyond 1 per AZ when limited to 36 SU

Yes these changes will keep a single ingress replica per az up to the connection demand of 36 SU.

> btw, this also assumes that this is blind to any throughput concerns.

Yes, that still has it's own calculation.

rareddy (approved):
lgtm

PR #838, opened 2022-11-22
MGDSTRM-10070 updating default connections to 9000

shawkins:


shawkins:
This is just for completeness.  I need to revisit this again and just remove all of the code that can setup the instance directly through the instance profile - any modifications that we want to do to the instance can be done through some other manipulation.

rareddy (approved):
lgtm

PR #839, opened 2022-12-02
PoC: Add an internal-to-OpenShift listener

grdryn:
JIRA: https://issues.redhat.com/browse/MGDSTRM-9651

This listener would be accessible through the <name>-kafka-brokers service, rather than through the ELB/Ingress controller setup.

The TLS cert is provided by the cluster CA, and client apps can be given access to that CA by following the docs here:

https://docs.openshift.com/container-platform/4.11/security/certificates/service-serving-certificate.html

One oddity is that the controller that creates the cert in response to the annotation on the Service will try to add an additional annotation also, but Strimzi removes that. I don't know if that causes any problems.


grdryn:
Closing this as it's not intended to be reviewed or merged (at least not for now).

PR #841, opened 2022-12-06
[MGDSTRM-10098] - Removing 9096 listener accessibility from other pods

biswassri:
Jira: [MGDSTRM-10098](https://issues.redhat.com/browse/MGDSTRM-10098)

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
@@ -1,5 +1,6 @@
 package org.bf2.operator.operands;
 
+import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.fabric8.kubernetes.api.model.networking.v1.NetworkPolicyPeerBuilder;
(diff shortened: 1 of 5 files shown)


operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
```diff
@@ -412,6 +413,11 @@ protected List<GenericKafkaListener> buildListeners(ManagedKafka managedKafka, i
                                 .withName("sre")
                                 .withPort(9096)
                                 .withType(KafkaListenerType.INTERNAL)
+                                .withNetworkPolicyPeers(new NetworkPolicyPeerBuilder()
+                                        .withIpBlock(new IPBlockBuilder()
+                                                .withCidr("0.0.0.0/32")
```
rareddy, line 418: It looks like it will block all IP addresses, how is 127.0.0.1 is open for 9096? is that implicit behavior?

MikeEdgar: This applies to traffic from outside the Pod via the Service. Loopback/local connections to 127.0.0.1 won't be affected.

rareddy (approved):
lgtm

vbusch (approved):
lgtm

shawkins (approved):
LGTM

PR #843, opened 2022-12-13
MGDSTRM-10277 hardening the sync against unhealthy behavior

shawkins:
This is purely trying to be more defensive to what was seen in MGDSTRM-10277.  However there remains a completely unknown root cause - I didn't see anything exceptional called out.  It's also not clear why only sync would be affected - we've moved up enough in the operator sdk such that it's using an informer for the primary controller resource (managedkafka) that should be exactly the same as what the sync is running.

### DIFF
--- a/sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
+++ b/sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
@@ -9,6 +9,7 @@ import io.fabric8.kubernetes.api.model.Secret;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.informers.cache.Cache;
+import io.fabric8.kubernetes.client.utils.KubernetesResourceUtil;
 import io.fabric8.kubernetes.client.utils.Serialization;
 import io.fabric8.zjsonpatch.JsonDiff;
 import io.micrometer.core.annotation.Counted;
@@ -18,6 +19,7 @@ import io.quarkus.scheduler.Scheduled.ConcurrentExecution;
 import org.bf2.common.ConditionUtils;
 import org.bf2.common.ManagedKafkaResourceClient;
 import org.bf2.common.OperandUtils;
+import org.bf2.common.health.SystemTerminator;
 import org.bf2.operator.ManagedKafkaKeys;
 import org.bf2.operator.resources.v1alpha1.ManagedKafka;
 import org.bf2.operator.resources.v1alpha1.ManagedKafkaCondition.Reason;
@@ -35,6 +37,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
 import java.net.HttpURLConnection;
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


shawkins:
Also this is not fully defensive - presumably because of missing events / invalid cache state the control plane was not receiving status updates either.  

If we end up seeing more frequent sync restarts due to these changes we'll have to find a root cause - and/or go back to using things like a timed restart job to ensure that the watches are fresh.

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -333,10 +345,18 @@ void create(ManagedKafka remote) {
...
-           log.infof("ManagedKafka %s already exists", Cache.metaNamespaceKeyFunc(remote));
+           // it's possible, but unlikely, that we'll see two reconcilitions close enough together
+           // such that we'll naturally see a conflict - that is the second completes before the cache
+           // is updated.  It's more likely that the cache state is not up-to-date.
+           exit(remote);
        }
     }
 
+    private void exit(ManagedKafka remote) {
+        log.errorf("ManagedKafka %s exists, but not in the informer cache.  The sync is not in a healthy state", Cache.metaNamespaceKeyFunc(remote));
```
MikeEdgar, line 356: It might be useful to generate a stack trace for the `errorf` to see which invocation brings us here.

MikeEdgar (approved):

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -333,10 +344,35 @@ void create(ManagedKafka remote) {
...
+     * Check if the cache state is consistent with the expectation that the entry does not exist in cache
+     * @return true if consistent
+     */
+    private boolean isConsistent(ManagedKafka remote) {
+        ManagedKafka existing = kubeClient.resources(ManagedKafka.class)
+                .inNamespace(remote.getMetadata().getNamespace())
+                .withName(remote.getMetadata().getName()).get();
+        if (existing != null) {
+            checkCreationTimestamp(existing, Duration.ofMinutes(1));
+            throw new RuntimeException(String.format("ManagedKafka %s exists, but not in the informer cache.  The sync may not be in a healthy state", Cache.metaNamespaceKeyFunc(remote)));
```
rareddy, line 365: Did you mean to use `InconsistentStateException` here?

shawkins: Sorry, that was an incomplete thought.  I was going to catch that exception for a create conflict, but it really is such a rare occurrence that we might as well let it turn into a full exception - if it is occurring with some regularity it would mean that the async reconcile tasks are being delayed then executed really close together.

Removed now.

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -105,7 +111,8 @@ public void syncKafkaClusters() {
             if (existing == null) {
                 if (!remoteSpec.isDeleted()) {
                     reconcileAsync(ControlPlane.managedKafkaKey(remoteManagedKafka), localKey);
-                } else {
+                } else if (isConsistent(remoteManagedKafka)) {
```
SamBarker, line 114: From the javadoc on `isConsistent` this condition makes sense but when reading the code it seems a bit strange to log a warning when something is consistent. I'm wondering if there is a way to rename `isConsistent` so that the usage reads better.

shawkins: This is a consistency check of local state.  The warning is logged because the remote state doesn't seem to match what we would expect.

rareddy (approved):
lgtm

PR #846, opened 2022-12-21
[MGDSTRM-10315] Restore simple ReconciliationPaused condition check for StrimziUpdating

MikeEdgar:
The changes to suspend a Kafka instance modified the `AbstractKafkaCluster#isReconciliationPaused` method to check for the presence of the Strimzi pause reconciliation annotation. Due to the amount of time this annotation is actually present during an upgrade, the `StrimziUpgrading` ready condition reason is set for a much shorter duration, impacting the `UpgradeST`. 

The actual upgrade of the Strimzi version isn't impacted, only the reporting of the ready reason, which is checked by the system test.

This PR also includes a change to the systemtests to align the Subscription label selection criteria more closely with those of the operator and also adds a check for the Strimzi versions in the agent status as a requirement for FSO readiness.


vbusch (approved):
lgtm

shawkins (approved):
LGTM

rareddy (approved):
lgtm

PR #847, opened 2023-01-05
MGDSTRM-10301 switching routes to use the loadbalancer hostname

shawkins:
This is to resolve the issue we saw with a private rosa cluster lacking a public entry in the DNS cluster config.  If that is the case, then the hostname of the ingress controller is only known within the private hosted zone - it cannot be used as a target for a route53 CNAME entry.  It is valid to instead directly use the loadbalancer hostname associated with the ingresscontroller.  This version of this change assumes we can do this all of time - but there are some concerns, such as we're effectively assuming that the loadbalancer hostname remains stable and that there's only one entry in the service status.  

Alternatives include:

- merging #840 first and only do this when managedkafkaagent is designated as public and we introspect the DNS config and see that lacks a public entry.  
- instead mandate that cluster owners create an appropriate public entry in their DNS config if they want public kafka.
- alleviate the concern about the loadbalancer host name stability by having the control plane update dns records - it looks like it only validates them currently.

This change is also blocked on the fleetmanager side in their DNS logic: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/db4c66b6510a4935f793ddad6dc60952fddb0ae6/internal/kafka/internal/services/data_plane_kafka.go#L515 there is an expectation that the router hostname has the same root as the cluster.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -6,6 +6,8 @@ import io.fabric8.kubernetes.api.model.ContainerBuilder;
 import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.LabelSelector;
+import io.fabric8.kubernetes.api.model.LoadBalancerIngress;
+import io.fabric8.kubernetes.api.model.LoadBalancerStatus;
 import io.fabric8.kubernetes.api.model.Node;
 import io.fabric8.kubernetes.api.model.NodeList;
 import io.fabric8.kubernetes.api.model.Pod;
@@ -14,6 +16,7 @@ import io.fabric8.kubernetes.api.model.Quantity;
 import io.fabric8.kubernetes.api.model.ResourceRequirements;
 import io.fabric8.kubernetes.api.model.ResourceRequirementsBuilder;
 import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.ServiceStatus;
 import io.fabric8.kubernetes.api.model.TolerationBuilder;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
@@ -111,12 +114,6 @@ public class IngressControllerManager {
     protected static final String INFRA_NODE_LABEL = "node-role.kubernetes.io/infra";
 
     protected static final String WORKER_NODE_LABEL = "node-role.kubernetes.io/worker";
-
-    /**
-     * Domain part prefixed to domain reported on IngressController status. The CNAME DNS records
-     * need to point to a sub-domain on the IngressController domain, so we just add this.
-     */
-    private static final String ROUTER_SUBDOMAIN = "ingresscontroller.";
 
     /**
      * Predicate that will return true if the input string looks like a broker resource name.
@@ -669,11 +666,13 @@ public class IngressControllerManager {
     }
 
     private String getIngressControllerDomain(String ingressControllerName) {
-        return ingressControllerInformer.getList().stream()
-                .filter(ic -> ic.getMetadata().getName().equals(ingressControllerName))
-                .map(ic -> ROUTER_SUBDOMAIN + (ic.getStatus() != null ? ic.getStatus().getDomain() : ic.getSpec().getDomain()))
-                .findFirst()
-                .orElse("");
+        return Optional.ofNullable(informerManager.getLocalService(INGRESS_ROUTER_NAMESPACE, "router-" + ingressControllerName))
+                .map(Service::getStatus)
+                .map(ServiceStatus::getLoadBalancer)
+                .map(LoadBalancerStatus::getIngress)
+                .flatMap(l -> l.stream().findFirst())
+                .map(LoadBalancerIngress::getHostname)
+                .orElse(null);
     }
 
     private Stream<Route> routesFor(ManagedKafka managedKafka) {
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -696,7 +696,7 @@ public class IngressControllerManager {
     }
 
     private String getIngressControllerDomain(String ingressControllerName, boolean hasPublicDns) {
-      if (!isPrivateNetwork(informerManager.getLocalAgent()) && !hasPublicDns) {
+      if (!hasPublicDns && !isPrivateNetwork(informerManager.getLocalAgent())) {
           // special case - nothing is creating the dns entries for the ingress controller, so we map to the load balancer
           return Optional.ofNullable(informerManager.getLocalService(INGRESS_ROUTER_NAMESPACE, "router-" + ingressControllerName))
               .map(Service::getStatus)
--- a/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
@@ -26,6 +26,8 @@ import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
 import io.fabric8.kubernetes.client.informers.cache.Cache;
 import io.fabric8.kubernetes.client.utils.CachedSingleThreadScheduler;
 import io.fabric8.kubernetes.client.utils.Serialization;
+import io.fabric8.openshift.api.model.DNS;
+import io.fabric8.openshift.api.model.DNSSpec;
 import io.fabric8.openshift.api.model.Route;
 import io.fabric8.openshift.api.model.operator.v1.ConfigBuilder;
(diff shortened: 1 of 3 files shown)


operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -656,11 +653,13 @@ private int numReplicasForConnectionDemand(double connectionDemand) {
...
-                .map(ic -> ROUTER_SUBDOMAIN + (ic.getStatus() != null ? ic.getStatus().getDomain() : ic.getSpec().getDomain()))
-                .findFirst()
-                .orElse("");
+        return Optional.ofNullable(informerManager.getLocalService(INGRESS_ROUTER_NAMESPACE, "router-" + ingressControllerName))
+                .map(Service::getStatus)
+                .map(ServiceStatus::getLoadBalancer)
+                .map(LoadBalancerStatus::getIngress)
+                .flatMap(l -> l.stream().findFirst())
+                .map(LoadBalancerIngress::getHostname)
+                .orElse(null);
```
MikeEdgar, line 662: I don't know if it's really necessary, but previously this method would never have returned `null`.
```suggestion
                .filter(Objects::nonNull)
                .orElse("");
```

MikeEdgar (approved):
This looks good to me. Regarding your comments on alternatives, I do think this approach is the cleanest/simplest in any case provided the domain name is stable.

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
shawkins: > I don't know if it's really necessary, but previously this method would never have returned null.

We can return empty, but it doesn't seem necessary - go clients in particular can't differentiate between empty and missing past deserialization.

MikeEdgar: > We can return empty, but it doesn't seem necessary

I'm fine with that, thanks

rareddy (approved):
lgtm

grdryn (approved):

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -109,12 +112,6 @@ public class IngressControllerManager {
 
     protected static final String WORKER_NODE_LABEL = "node-role.kubernetes.io/worker";
 
-    /**
-     * Domain part prefixed to domain reported on IngressController status. The CNAME DNS records
-     * need to point to a sub-domain on the IngressController domain, so we just add this.
-     */
-    private static final String ROUTER_SUBDOMAIN = "ingresscontroller.";
```
machi1990, line 116: Will this change affect existing Kafkas somehow and a migration will be needed?

machi1990: @shawkins

shawkins: As coded yes we would need both the check the control plane is doing relaxed and the control plane logic to update cname records implemented.  We'll leave the do not merge label on this until we get both of these points straightened out.

rareddy: what does mean by "control plane is doing relaxed"? Also since the URL based on customer's load balancer, why should register cname in RHOSAK?

machi1990: > what does mean by "control plane is doing relaxed"?

The control plane currently validates that the returned route URL by the fleet sync corresponds to the OSD cluster dns: https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/db4c66b6510a4935f793ddad6dc60952fddb0ae6/internal/kafka/internal/services/data_plane_kafka.go#L515

The OSD cluster dns is retrieved from cluster ingresses. I understand this won't be the case for private cluster and the check will have to be relaxed somehow. 

> Also since the URL based on customer's load balancer, why should register cname in RHOSAK?

If the route URL changes, KFM needs to update the cname record. From what I see, the existing router all had a `ingresscontroller.` prefix. If this URL is changed somehow, the CNAMEs record will need to be updated to point to the new routes.

rareddy: Thanks @machi1990 

On CNAME registration, what I mean is since the URL is backed by users' AWS Route 53 already (I am assuming as it is their load balancer) why also registering in RHOSAK's AWS route 53.

shawkins: @rareddy the scenario is a public kafka on an cluster that lacks a public dns entry in their config - we have already seen that this is the default state of private ROSA clusters.  If we are in that circumstance, creating cname entries that point from the application domain endpoints broker1.something to ingresscontroller.kas-zone.clusterdomain will not work - as nothing is creating a public dns entry for kas-zone.clusterdomain.  The necessary entry will point from kas-zone.clusterdomain to the aws loadbalancer hostname.  So here we are skipping that and pointing broker1.something directly to aws loabalancer hostname.  A way to do this that would address the migration concern would be leave the top level dns entries as is - and create additional route entries to point from ingresscontroller.kas-zone.clusterdomain to the aws loadbalancer hostname.  That will still require a change on the control plane side however as the target still won't be based upon the cluster domain.

shawkins: No the ingresscontroller is still used.  We still are routing traffic to the aws loadbalancer.  The aws loadbalancer has already been configured by the ingress controller operator to point to the ingress controller service / ha proxy pods.  

We can only get away from using the ingress controllers if we use loadbalancer services or switch to using ingress and the https://github.com/kubernetes-sigs/aws-load-balancer-controller - I didn't see that the operator for this is running by default, let me do some more testing on a local instance.

shawkins:
The next changes make the switch to the loadbalancer hostname condition on it being a public kafka on a cluster that lacks a public dns config.  This addresses an immediate need to worry about migrating / modifying existing dns entries - my understanding in that case was that we'd still continue to function appropriately as the existing entries would still be in place but we'd fail the validation that checks if the cname records need to change.

The other changes in the operator config - @rareddy was it intentional to leave in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/pull/847/files#diff-0a4c59e14630b9f89b740bf0d391beda94261fcf66093ddba7d9226761e90960L227 - I didn't think we actually wanted it hard-coded in this manner, so it's been removed.

I'm also overriding the fabric8 max concurrent values - we're working to address this in 6.3+, but right now if you create 20 websockets then okhttp won't submit any additional requests.  With the new dns informer, I count 18 informers we're creating and the operator sdk should be creating another - dangerously close to the default limit.  That would also force anything that we have as concurrent logic to be forced over a limited number of connections.

machi1990 (approved):
lgtm, thanks for adding in the code change to require no migration of already existing dns entries
@shawkins

operator/src/main/kubernetes/kubernetes.yml
```diff
@@ -224,8 +224,10 @@ spec:
           env:
             - name: QUARKUS_PROFILE
               value: prod
-            - name: INGRESSCONTROLLER_AZ_REPLICA_COUNT
-              value: "1"              
+            - name: KUBERNETES_MAX_CONCURRENT_REQUESTS
+              value: 100
+            - name: KUBERNETES_MAX_CONCURRENT_REQUESTS_PER_HOST
+              value: 100
```
MikeEdgar, line 230: Why not put this in `application.properties`?

operator/src/main/java/org/bf2/operator/managers/IngressControllerManager.java
```diff
@@ -668,12 +695,31 @@ private int numReplicasForConnectionDemand(double connectionDemand) {
...
     }
 
-    private String getIngressControllerDomain(String ingressControllerName) {
-        return ingressControllerInformer.getList().stream()
-                .filter(ic -> ic.getMetadata().getName().equals(ingressControllerName))
-                .map(ic -> ROUTER_SUBDOMAIN + (ic.getStatus() != null ? ic.getStatus().getDomain() : ic.getSpec().getDomain()))
-                .findFirst()
-                .orElse("");
+    private String getIngressControllerDomain(String ingressControllerName, boolean hasPublicDns) {
+      if (!isPrivateNetwork(informerManager.getLocalAgent()) && !hasPublicDns) {
```
MikeEdgar, line 699: Optional efficiency nit - 
```suggestion
      if (!hasPublicDns && !isPrivateNetwork(informerManager.getLocalAgent())) {
```

MikeEdgar (approved):
LGTM, but a couple comments.

operator/src/main/kubernetes/kubernetes.yml
MikeEdgar: Got it, thanks - I didn't realize these weren't the typical Quarkus/MP configs.

shawkins: I'm not sure why, but quarkus does not provide config hooks for these.  Seems like it would have become an issue for other pre-fabric8 6.4 users.

rareddy (approved):
lgtm

PR #851, opened 2023-01-17
MGDSTRM-10325: Delete Strimzi cluster operator leadership lease on deployment deletion

robobario:
Why:
As part of enabling StrimziPodSets in the cluster operator we are also increasing availability of the operator by running 2 replicas and using a k8s lease to elect one of them as a leader, the other will be in standby in case it needs to take over reconciliation. The lease is not able to be automatically cleaned up when we tear down a cluster operator deployment, so we are making the Fleetshard Operator responsible for deleting the lease in reaction to a deployment deletion.

This is the upstream issue in the fabric8 client https://github.com/fabric8io/kubernetes-client/issues/4638, once this is resolved we can remove this work from Fleetshard and configure an owner reference in the Strimzi cluster operator

Note that the lease is named ${deploymentName}-leadership-token by convention to enable fleetshard to clean it.

https://issues.redhat.com/browse/MGDSTRM-10325

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/StrimziManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/StrimziManager.java
@@ -5,8 +5,10 @@ import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.PodTemplateSpec;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.api.model.apps.DeploymentList;
+import io.fabric8.kubernetes.api.model.coordination.v1.Lease;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
+import io.fabric8.kubernetes.client.dsl.Resource;
 import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
 import io.fabric8.kubernetes.client.internal.readiness.Readiness;
 import io.fabric8.kubernetes.client.utils.Serialization;
@@ -128,10 +130,29 @@ public class StrimziManager {
                             log.debugf("Delete event received for Deployment %s/%s",
                                     deployment.getMetadata().getNamespace(), deployment.getMetadata().getName());
                             deleteStrimziVersion(deployment);
+                            deleteLeadershipLeaseIfPresent(deployment);
                             updateStatus();
                         }
                     }
                 });
+    }
+
+    void deleteLeadershipLeaseIfPresent(final Deployment deployment) {
+        final String leaseName = deployment.getMetadata().getName() + "-leadership-token";
+        final String namespace = deployment.getMetadata().getNamespace();
+        try {
+            log.debugf("Deleting lease %s in namespace %s", leaseName, namespace);
+            final Resource<Lease> lease = kubernetesClient.leases().inNamespace(namespace)
+                                                          .withName(leaseName);
+            final Boolean success = lease.delete();
+            if(success == null || !success){
+                log.warnf("Strimzi cluster-operator leadership lease %s was not found in namespace %s when attempting deletion", leaseName, namespace);
+            } else {
+                log.debugf("Strimzi cluster-operator leadership lease %s successfully deleted from namespace %s", leaseName, namespace);
+            }
+        } catch (Exception e){
+            log.errorf("Failed to delete Strimzi cluster-operator leadership lease %s in namespace %s", e, leaseName, namespace);
+        }
     }
 
     private boolean isStrimziDeployment(Deployment deployment) {
--- a/operator/src/main/java/org/bf2/operator/managers/StrimziManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/StrimziManager.java
@@ -146,7 +146,7 @@ public class StrimziManager {
      * <p>
      * See <a href="https://github.com/fabric8io/kubernetes-client/issues/4638">fabric8 kubernetes-client issue #4638</a>
      */
-    void deleteLeadershipLeaseIfPresent(final Deployment deployment) {
+    private void deleteLeadershipLeaseIfPresent(final Deployment deployment) {
         final String leaseName = deployment.getMetadata().getName() + "-leadership-token";
         final String namespace = deployment.getMetadata().getNamespace();
         try {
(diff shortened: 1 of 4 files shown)


operator/src/test/java/org/bf2/operator/managers/StrimziManagerTest.java
```diff
@@ -205,6 +211,30 @@ public void testUninstallingStrimziOperator() {
...
+                .endMetadata()
+                .withNewSpec()
+                    .withHolderIdentity("me")
+                    .withLeaseDurationSeconds(1000)
+                    .withAcquireTime(ZonedDateTime.ofInstant(Instant.now(), ZoneId.of("UTC")))
+                .endSpec().build();
+        server.getClient().leases().inNamespace("ns-1").create(lease);
+        installStrimziOperator("strimzi-cluster-operator.v1", "ns-1", "2.7.0=kafka-2.7.0", true, true);
+        Lease actual = server.getClient().leases().withName("strimzi-cluster-operator.v1-leadership-token").get();
+        assertNull(actual);
```
robobario, line 229: realized the test is incorrect here, also I need to figure out how to trigger the onDelete watcher since the uninstallStrimziOperator doesn't trigger it as hoped

operator/src/test/java/org/bf2/operator/managers/StrimziManagerTest.java
```diff
@@ -58,10 +66,11 @@ public class StrimziManagerTest {
 
     @BeforeEach
     public void beforeEach() {
-        // before each test clean Kubernetes server (no Deployments from other runs)
-        this.server.before();
+        try (NamespacedKubernetesClient client = this.server.getClient().inAnyNamespace()) {
```
robobario, line 69: Without this change the test was failing because the StrimziManagerTest instance and the StrimziManager instance had clients pointing at different mock kubernetes servers on different ports. The test created a lease in one mock server and the implementation tried to delete it from another which was a no-op.

operator/src/test/java/org/bf2/operator/managers/StrimziManagerTest.java
```diff
@@ -205,6 +214,30 @@ public void testUninstallingStrimziOperator() {
...
+                    .withAcquireTime(ZonedDateTime.ofInstant(Instant.now(), ZoneId.of("UTC")))
+                .endSpec().build();
+        server.getClient().leases().inNamespace("ns-1").create(lease);
+        final Deployment deployment = installStrimziOperator("strimzi-cluster-operator.v1", "ns-1", "2.7.0=kafka-2.7.0", true, true);
+        final Resource<Lease> leaseResource = server.getClient().leases().inNamespace("ns-1").withName("strimzi-cluster-operator.v1-leadership-token");
+        Lease actual = leaseResource.get();
+        assertNotNull(actual);
+        List<StrimziVersionStatus> strimziVersions = this.strimziManager.getStrimziVersions();
+        assertTrue(checkStrimziVersion(strimziVersions, "strimzi-cluster-operator.v1", true));
+        strimziManager.deleteLeadershipLeaseIfPresent(deployment);
```
robobario, line 236: I wasn't sure how to invoke the ResourceEventHandler which the test passes into a `MockResourceInformerFactory`, maybe with some re-working we could keep a reference somewhere in the mock. Maybe a future improvement to make them testable.

katheris (approved):
LGTM

Only note is it looks like the delete DSL in the fabric8 client has a different return type in the latest version, so when kas-fleetshard upgrades their fabric8 client the code will likely need a slight change, but for now it looks good to me.

operator/src/test/java/org/bf2/operator/managers/StrimziManagerTest.java
k-wall: that sounds like we've got really fragile test code.  @shawkins are you aware of this problem?  Is there a nicer way?

operator/src/main/java/org/bf2/operator/managers/StrimziManager.java
```diff
@@ -128,12 +130,31 @@ public void onDelete(Deployment deployment, boolean deletedFinalStateUnknown) {
...
                             deleteStrimziVersion(deployment);
+                            deleteLeadershipLeaseIfPresent(deployment);
                             updateStatus();
                         }
                     }
                 });
     }
 
+    void deleteLeadershipLeaseIfPresent(final Deployment deployment) {
+        final String leaseName = deployment.getMetadata().getName() + "-leadership-token";
```
k-wall, line 141:
```suggestion
        // See https://github.com/fabric8io/kubernetes-client/issues/4638
        final String leaseName = deployment.getMetadata().getName() + "-leadership-token";
```

k-wall:
@robobario the approach looks good to me.

shawkins:
> Only note is it looks like the delete DSL in the fabric8 client has a different return type in the latest version, so when kas-fleetshard upgrades their fabric8 client the code will likely need a slight change, but for now it looks good to me.

Yes - the boolean return value isn't generally useful and confusing for some of the multi-delete scenarios.  Later versions return a list of items marked for deletion and the ability to await all of those deletions to be processed.

operator/src/test/java/org/bf2/operator/managers/StrimziManagerTest.java
shawkins: We don't want to call before.  Until we get to a later version of fabric8, you mostly need to clean the state of the mock server yourself in-between tests.  Due to a quarkus / okhttp issue we cannot scope the mock server to test classes - it is scoped to the whole module.  In fabric8 6.x there is a KuberentesMockServer reset method that should make things easy regardless.

operator/src/main/java/org/bf2/operator/managers/StrimziManager.java
robobario: Added it as javadoc to the method, definitely worth a comment for this oddity

k-wall (approved):
lgtm

operator/src/test/java/org/bf2/operator/managers/StrimziManagerTest.java
robobario: I went and had a shot at this, keeping references to all handlers created per type so that the test can access and call them directly. Maybe it would be better if `StrimziManager` implemented `ResourceEventHandler<Deployment>` and we didn't us an anonymous inner class, then we could call the methods directly on StrimziManager.

PR #854, opened 2023-01-17
[MGDSTRM-9790] Test instance upgrade while suspended

MikeEdgar:
- Add test for MGDSTRM-9790 to suspend, upgrade Strimzi, and resume an MK
- Increase operator reconcile max retry attempts, used when unhandled exceptions occur
- Delay sync's polling of mock control plane in test profile (account for observed HTTP server startup time)
- Modify sync's MK agent informer to only watch own namespace (aligns with `ManagedKafkaAgentSync` processing)
- Deploy Keycloak and kas-fleetshard only once per suite execution

### DIFF
--- a/systemtest/src/test/java/org/bf2/systemtest/unit/SuiteUnitTest.java
+++ b/systemtest/src/test/java/org/bf2/systemtest/unit/SuiteUnitTest.java
@@ -14,16 +14,23 @@ import io.quarkus.test.kubernetes.client.KubernetesServerTestResource;
 import io.quarkus.test.kubernetes.client.KubernetesTestServer;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.bf2.systemtest.framework.ExtensionContextParameterResolver;
+import org.bf2.systemtest.framework.IndicativeSentences;
 import org.bf2.systemtest.framework.KeycloakInstance;
 import org.bf2.systemtest.framework.ParallelTest;
 import org.bf2.systemtest.framework.SecurityUtils;
 import org.bf2.systemtest.framework.SequentialTest;
-import org.bf2.systemtest.integration.AbstractST;
+import org.bf2.systemtest.framework.TestCallbackListener;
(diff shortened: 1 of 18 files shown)


MikeEdgar:
While trying to add a fourth test without impacting the duration of the tests too much, this PR certainly grew larger than what was expected for MGDSTRM-9790. I'm happy to split out the other changes if it is easier to review that way.

test/src/main/java/org/bf2/test/TestUtils.java
```diff
@@ -165,16 +166,29 @@ public static Function<InputStream, InputStream> replacer(final Map<String, Stri
      * Restart kubeapi
      */
     public static void restartKubeApi() {
```
shawkins, line 168: I'd be fine with removing this method and the test that is based upon it.  This will not cause an informer to relist, just the underlying watches to reconnect.  We also no longer have our own informer logic.

systemtest/src/test/java/org/bf2/systemtest/unit/SuiteUnitTest.java
```diff
@@ -42,9 +49,16 @@
  */
 @QuarkusTest
 @QuarkusTestResource(KubernetesServerTestResource.class)
-public class SuiteUnitTest extends AbstractST {
```
shawkins, line 45: What was the motivation to remove the base class?

MikeEdgar: Aside from the annotations, the only thing gained was the setup of the `KubeClient`. I figured the three other *ST classes had more to gain by consolidating setup in the base class without adding an intermediate parent.

MikeEdgar:
@shawkins please let me if you get back to this and have any other comments.

@kornys please also take a look if you have time next week.

I can split this up if it's too much to consume for one PR.

shawkins (approved):
> @shawkins please let me if you get back to this and have any other comments.

No other comments, the application code changes are minimal and I'm fine with the test changes.

kornys (approved):
LGTM!

PR #850, opened 2023-01-13
MGDSTRM-9181: Have kas ingress controller utilise the openshift-router's dynamic config manager.

k-wall:
Enables the (unsupported) dynamic config manager option on openshift-ingress.  This is done as otherwise, haproxy reconfigurations (which occur every time new kafka instance is provisioned or deprovisioned) cause all kafka connections to all instances to be dropped.  This is too disruptive for applications.

Work based on ideas from @shawkins and @showuon .

PR #856, opened 2023-01-19
bug fix: create blueprints for suspended routes too

k-wall:
#850 was incomplete as we failed to consider suspended routes, so instance suspension was still leading to a haproxy restart.  This creates blueprints for those routes too.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
@@ -333,7 +333,11 @@ public abstract class AbstractKafkaCluster implements Operand<ManagedKafka> {
                 routeAnnotations.putAll(rateLimitAnnotations);
                 r.getMetadata().setAnnotations(routeAnnotations);
                 OperandUtils.createOrUpdate(kubernetesClient.resources(Route.class), r);
+                ensureBlueprintRouteForSuspendedRoute(r);
             });
+    }
+
+    protected void ensureBlueprintRouteForSuspendedRoute(Route r) {
     }
 
     protected List<GenericKafkaListener> buildListeners(ManagedKafka managedKafka, int replicas) {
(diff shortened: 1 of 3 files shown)


shawkins (approved):
LGTM

grdryn (approved):

k-wall:
I've confirmed this is properly handling the suspended case.  Suspending an instance results in the addition of two more blueprints.

The bootstrap one has:

```
    haproxy.router.openshift.io/balance: leastconn
    haproxy.router.openshift.io/rate-limit-connections: "true"
    haproxy.router.openshift.io/rate-limit-connections.concurrent-tcp: "5"
    haproxy.router.openshift.io/rate-limit-connections.rate-tcp: "15"
```
and the one corresponding to the normal broker routes:

```
   haproxy.router.openshift.io/rate-limit-connections: "true"
    haproxy.router.openshift.io/rate-limit-connections.concurrent-tcp: "5"
    haproxy.router.openshift.io/rate-limit-connections.rate-tcp: "15"
```

Looking at memory, the ingress pods with the extra blueprints are now sitting at ~350Mi.

shawkins:
> Looking at memory, the ingress pods with the extra blueprints are now sitting at ~350Mi.

Presumably the 500 setting is per pool, so I think we could safely drop that to 300 (50 instances * 3 brokers * 2 margin) now that we're subdividing things so much.

k-wall:
> > Looking at memory, the ingress pods with the extra blueprints are now sitting at ~350Mi.
> 
> Presumably the 500 setting is per pool, so I think we could safely drop that to 300 (50 instances * 3 brokers * 2 margin) now that we're subdividing things so much.

Yes, 500 is per pool.  I'd be happy with 300 pool size.   I'll do it.  As we mentioned elsewhere there is provision to annotate the blueprint with a poolsize (after all 300 suspended instances = unlikely), but I vote we don't do that now.  We can tune that later if we like.

PR #859, opened 2023-01-19
MGDSTRM-10459 changing the reloadInterval tuning option to a string

shawkins:
We need to leave the unsupport option as an int, but change the tuning option to a string.


rareddy (approved):
lgtm

I am assuming this works in versions 4.10 and above? I mean string conversion is not specific to 4.12+

MikeEdgar (approved):

k-wall (approved):

shawkins:
> I am assuming this works in versions 4.10 and above? I mean string conversion is not specific to 4.12+

The failure is due to the compatibility logic that I had already put in.  In both 4.11 and 4.12 the openshift cluster ingress operator moved properties from unsupported to tuning options.  In doing so if you simply left the cr with the unsupported option, that feature will no longer be enabled.  So the logic rather than performing a version check is setting the properties in both tuning options and unsupported properties.  If the tuning option does not exist on 4.10 or 4.11, it is simply dropped by the api server.

PR #863, opened 2023-01-31
MGDSTRM-9751 - JOSDK and fabric8 update

biswassri:


operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaAgentController.java
```diff
@@ -44,9 +42,7 @@
  */
 @ApplicationScoped
 @ControllerConfiguration(
-        finalizerName = Constants.NO_FINALIZER,
```
biswassri, line 47: @MikeEdgar The updated operator sdk doesn't have the same constants. Are we using this `finalizerName` value somewhere else?

MikeEdgar: Remove it - it's not needed in the newer version. The docs say it will only add a finalizer if the `Cleaner` interface is implemented by the controller.

operator/src/main/java/org/bf2/operator/events/ResourceEventSource.java
```diff
@@ -45,7 +45,7 @@ protected void handleEvent(HasMetadata resource, ResourceAction action) {
         // the operator may not have inited yet
         if (getEventHandler() != null) {
             ResourceID.fromFirstOwnerReference(resource).ifPresentOrElse(
-                    ownerId -> getEventHandler().handleEvent(new ResourceEvent(action, ownerId)),
```
biswassri, line 48: @MikeEdgar  this is the only change that I haven't been able to resolve. Can you help me with this?

MikeEdgar: Here you can pass `null` as the third parameter.

operator/src/test/java/org/bf2/operator/events/ControllerEventFilterTest.java
```diff
@@ -27,7 +27,7 @@ void setup() {
 
     @Test
     void testAddResourceAccepted() {
-        assertTrue(target.acceptChange(mockConfiguration, null, new ManagedKafka()));
+        assertTrue(target.accept(null, new ManagedKafka()));
```
MikeEdgar, line 30: This test method and the `mockConfiguration` field and code in `setup` can also be removed. We no longer need to test the `add` case because the filter is only for updates.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaController.java
```diff
@@ -41,17 +35,14 @@
 import javax.inject.Inject;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
 @ControllerConfiguration(
```
MikeEdgar, line 43: Also add the `ControllerEventFilter.class` as the `onUpdateFilter` on this annotation.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaAgentController.java
biswassri: Sure. It's already removed!

operator/src/test/java/org/bf2/operator/events/ControllerEventFilterTest.java
biswassri: Sure. I'll update this.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaController.java
```diff
@@ -128,10 +121,6 @@ public UpdateControl<ManagedKafka> reconcile(ManagedKafka managedKafka, Context
         }
     }
 
-    @Override
-    public List<EventSource> prepareEventSources(EventSourceContext<HasMetadata> context) {
-        return Arrays.asList(eventSource);
-    }
```
MikeEdgar, line 134: This and the `EventSourceInitializer<HasMetadata>` interface are still needed. It looks like the return type changed to `Map<String, EventSource>` so the method body could instead be something like `return Map.of("ownedResources", eventSource)`

MikeEdgar (changes requested):
Just one more thing - the controller still needs to provide the event source.

operator/src/main/java/org/bf2/operator/controllers/ManagedKafkaController.java
biswassri: Added this back!

MikeEdgar (approved):
lgtm :+1:

rareddy (approved):
lgtm

biswassri:
Tested out the changes, created KI and smoke test successfully.

shawkins (approved):
LGTM

PR #865, opened 2023-02-09
WIP: Attempting to create a new perf test

grdryn:


perf/src/main/java/org/bf2/performance/OMB.java
```diff
@@ -266,6 +275,7 @@ private void createWorker(String jvmOpts, String name, Node node) {
                 .withName("ca")
                 .editOrNewSecret()
                 .withSecretName("ext-listener-crt")
+                .withOptional(true)
```
grdryn, line 278: When targeting an instance that has a trusted cert, there won't be a need for a truststore.

PR #866, opened 2023-02-10
[MGDSTRM-10669] Allow access via 9096 for any Kafka pod in namespace

MikeEdgar:
Test OLM index image with this change: `quay.io/medgar/kas-fleetshard-operator-index:1.0.17-1`

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
@@ -2,7 +2,6 @@ package org.bf2.operator.operands;
 
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.Quantity;
-import io.fabric8.kubernetes.api.model.networking.v1.IPBlockBuilder;
 import io.fabric8.kubernetes.api.model.networking.v1.NetworkPolicyPeerBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.openshift.api.model.Route;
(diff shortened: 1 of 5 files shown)


biswassri (approved):
lgtm

k-wall (approved):

operator/src/main/java/org/bf2/operator/operands/AbstractKafkaCluster.java
```diff
@@ -417,17 +416,18 @@ protected List<GenericKafkaListener> buildListeners(ManagedKafka managedKafka, i
...
                         new GenericKafkaListenerBuilder()
                                 .withName("sre")
                                 .withPort(9096)
                                 .withType(KafkaListenerType.INTERNAL)
                                 .withNetworkPolicyPeers(new NetworkPolicyPeerBuilder()
-                                        .withIpBlock(new IPBlockBuilder()
-                                                .withCidr("0.0.0.0/32")
-                                                .build())
+                                        .withNewPodSelector()
+                                            .addToMatchLabels("strimzi.io/name", managedKafka.getMetadata().getName() + "-kafka")
```
k-wall, line 429: Don't we need a namespace selector too? Otherwise an evil person could label any old pod with that label and have access?

https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name

MikeEdgar: I thought that as well, but I think this applies [1]:

> podSelector: This selects particular Pods in the same namespace as the NetworkPolicy which should be allowed as ingress sources or egress destinations.

[1] https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

MikeEdgar: Nonetheless. this thread is making me question it so I'll try that :-)

k-wall (approved):

k-wall:
I tested both negative and positive cases on a kafka instance with 3 brokers.  All good now.  Thanks @MikeEdgar

PR #870, opened 2023-02-16
MGDSTRM-10094 making KafkaCluster.fromKafka not modify the ManagedKafka

shawkins:
This moves the upgrade annotations from the managedkafka to kafka.  After this version is rolled out, we'll be able to remove the logic in the controller that is checking for moving the annotations and remove the annotations from the list of ones that are managed by the operator.


MikeEdgar (approved):
LGTM. I haven't been able to try this fully since we don't have a recent Strimzi bundle with multiple Kafka versions. I've opened MGDSTRM-10779 to address that.

biswassri (approved):
LGTM

shawkins:
@ppatierno if there are no objections, I'll go ahead and merge this.

ppatierno:
No objections ;-)

PR #869, opened 2023-02-16
[MGDSTRM-10711] Add operator/sync "high" priority class

MikeEdgar:
Adding `kas-fleetshard-high` now that the the other priority classes are present.

PR #871, opened 2023-02-20
[MGDSTRM-10719] Add build/push of bundle and index images

MikeEdgar:
Builds on #869 .


operator/src/main/java/org/bf2/operator/KasFleetShardOperator.java
```diff
@@ -13,7 +17,23 @@
...
 import java.util.stream.StreamSupport;
 
-public class KasFleetShardOperator implements QuarkusApplication {
+@CSVMetadata(
+        description = "Operator That Manages Kafka Instances",
+        displayName = "KaaS Fleetshard Operator",
+        provider = @Provider(
+                name = "bf2 community",
+                url = "https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard"),
+        minKubeVersion = "1.21.0",
```
MikeEdgar, line 26: I suspect it's higher than 1.21 by now. We could bump to 1.24 to align with OCP 4.11

MikeEdgar:
Coverage will not display properly for the new module until the `sonar.yml` changes are on `main`. I will add a README for the new module, but this is otherwise ready for review. Most of the new files are test resources for the unit test.

pom.xml
```diff
@@ -92,6 +92,7 @@
 
     <modules>
         <module>api</module>
+        <module>bundle</module>
```
shawkins, line 95: Should the bundle module always be active?

bundle/src/main/java/org/bf2/kas/fleetshard/bundle/BundleAssembler.java
```diff
@@ -0,0 +1,526 @@
...
+            })
+            // Sorted to ensure CRDs are in the same order for testing/comparison
+            .sorted(Comparator.<Map<?, ?>, String>comparing(crdExample -> crdExample.get("kind").toString())
+                    .thenComparing(crdExample -> crdExample.get("apiVersion").toString()))
+            .collect(Collectors.toList());
+    }
+
+    void copyPriorityClasses() throws IOException {
+        try (KubernetesClient client = new DefaultKubernetesClient()) {
+            client.load(operatorBundle.getEntry("kubernetes.yml")).get().stream()
```
shawkins, line 244: Just a heads up, later versions of fabric8 require using .items() instead of get() for this.

pom.xml
MikeEdgar: I figured yes for the build and unit tests, but the actual bundle generation will only happen when the `generate-bundle` is active.

shawkins (approved):
LGTM

biswassri (approved):
Lgtm. Just the $USER change that we discussed. Everything else worked fine for me.

bundle/README.md
```diff
@@ -0,0 +1,46 @@
...
+    "path": "/spec/install/spec/deployments/1/spec/template/spec/imagePullSecrets",
+    "value": [{ "name": "kas-image-pull-secret" }]
+  }]' \
+  -Dkas.bundle.credential.docker-config-path=/run/user/$(id -u)/containers/auth.json \
+  -Dkas.bundle.image=quay.io/${QUAY_USERNAME}/kas-fleetshard-operator-bundle:latest \
+  -Dkas.index.image-registry=quay.io \
+  -Dkas.index.image-group=${QUAY_USERNAME} \
+  -Dkas.index.image=kas-fleetshard-operator-index \
+  -Dkas.index.image-tag=latest \
+  -Dkas.index.build-engine=podman
```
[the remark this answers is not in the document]
kornys: we can also use buildah as we use that in jenkins on ocp pods

kornys (approved):

MikeEdgar:
Updated the documentation with suggestions and reverted the changes to systemtests/pom.xml. That will be done instead in a separate PR. I'll merge once the checks are green (aside from the coverage).

PR #873, opened 2023-03-01
[MGDSTRM-10764] Update Keycloak to version 21 for OpenShift 4.12

MikeEdgar:
- Bump Keycloak to version 21, adjust installation and supporting test code
- Specify JUL log manager property for system tests - Quarkus 2.13.7 upgrade introduced very verbose OkHttp logging at INFO level, which is limited with this property
- Remove surefire configuration to allow unit tests to be skipped when running with the `quickly` profile


systemtest/src/main/resources/keycloak-postgres.yml
```diff
@@ -0,0 +1,54 @@
...
+metadata:
+  name: keycloak-postgres
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+---
+kind: StatefulSet
```
kornys, line 13: just question, why statefulset? I think deployment is enough or not? :)

MikeEdgar: I'd say that StatefulSets are typical for databases. Since there is only a single replica here it probably doesn't matter, but better to follow common practice and the Keycloak docs[1]. 

[1] https://www.keycloak.org/operator/basic-deployment

kornys (approved):
Have no chance to test it, but LGTM from code POV

MikeEdgar:
@biswassri @shawkins ok to merge this? The changes are entirely in the `systemtest` module.

shawkins (approved):
LGTM

systemtest/src/main/java/org/bf2/systemtest/operator/KeycloakOperatorManager.java
```diff
@@ -1,104 +1,205 @@
...
     }
 
+    static void createOrReplace(KubernetesClient client, HasMetadata resource) {
+        resource.getMetadata().setNamespace(OPERATOR_NS);
+        client.resource(resource).inNamespace(OPERATOR_NS).createOrReplace();
+        INSTALLED_RESOURCES.add(resource);
+    }
+
+    static void awaitPodReady(String description, String labelName, String labelValue) {
+        TestUtils.waitFor(description + " ready", 1_000, 600_000, () -> isPodReady(labelName, labelValue));
```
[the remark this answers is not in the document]
MikeEdgar: Thanks, good suggestion. I opened #876 so we can remember to revisit this.

PR #874, opened 2023-03-01
Use OLM bundle image digest file to pin image used by index

MikeEdgar:
Allows the bundle image to be referenced via SHA rather than tag by the index

Relates to MGDSTRM-10595


shawkins (approved):
LGTM

biswassri (approved):
Lgtm

PR #877, opened 2023-03-06
MGDSTRM-10828 updating github actions

shawkins:
Updating various action versions.  Moving caching to the setup-java action.  Having everything use temurin - previously zulu (the default), adopt, and temurin were referenced.  Not sure if we have a hard preference here.


shawkins:
> LGTM. There is a `v4` for actions/labeler in auto-labeler.yaml.

Indeed, updated that one as well.

kornys (approved):

biswassri (approved):
LGTM

PR #879, opened 2023-03-09
[MGDSTRM-10786] Disable java-operator-sdk controller metrics by default

MikeEdgar:


rareddy:
what is the effect of this on any metrics that we may be capturing from this endpoint for management purposes?

MikeEdgar:
> what is the effect of this on any metrics that we may be capturing from this endpoint for management purposes?

The operator will no longer store or return metrics in the `operator_sdk_` prefix. I haven't found anywhere they are actually used for anything currently.

rareddy (approved):
lgtm

biswassri (approved):
LGTM

shawkins (approved):
LGTM

PR #881, opened 2023-03-13
MGDSTRM-10828 more github action updates

shawkins:


kornys (approved):

.github/workflows/sonar.yml
```diff
@@ -67,17 +65,6 @@ jobs:
         with:
           path: ./target/build-context.json
 
-      ## (PRs Only) Check out the base branch (target of the PR)
```
MikeEdgar, line 70: I think this is still needed. Without it, the workflow runs with `main`'s code and the PR's `target` output.

MikeEdgar: I take that back, this check's out the target. So having it here allows for PRs that target branches other than `main`

MikeEdgar:
The error in the job looks familiar, I think you might need to add `.rest` in the method invocation that fetches the remote build results.

https://github.com/smallrye/smallrye-open-api/commit/a19098a0129c3481ec413138841d7dbd2ff48f9a#diff-9a1c250ec072fc76ad44435cd9462d5693ea81f304d650922d3c2792bef267c2

shawkins:
> The error in the job looks familiar, I think you might need to add .rest in the method invocation that fetches the remote build results.

Yes, that is missing.

.github/workflows/sonar.yml
shawkins: To clarify this seems to be trying to run the sonar analysis post merge, is that correct?  I'm not sure why we're only doing that in this workflow.

MikeEdgar: When it's a PR analysis (versus an analysis after the PR is merged), It sets up the local git with the source of the PR (which might be a fork) and here it adds a separate `upstream` remote that Sonar uses to calculate the PR's deltas (e.g. for new bugs added with the PR).

If it's just a push to main, the analysis doesn't include any of the delta information and the secondary checkout isn't needed.

The repo/ref params in the `Checkout` step are needed to support that.

shawkins: What confused me was why this was needed - based upon https://docs.sonarqube.org/latest/analyzing-source-code/pull-request-analysis/ it says this is done automatically.  As well as not being necessary for fabric8 https://github.com/fabric8io/kubernetes-client/blob/master/.github/workflows/sonar.yml

MikeEdgar: I'm trying to figure out how fabric8 is setup. What I've always seen in the past is that the Sonar scan fails because it runs with the fork's secrets. This is (was?) a security feature of Github to prevent untrusted code from having access to the target repo's secrets.

shawkins: Looking at https://docs.sonarqube.org/9.6/devops-platform-integration/github-integration/ shows a pretty straight-forward setup in the maven example as well.  I've reverted these modifications for now as they were obviously not done completely.  Then we can determine if any further changes are warranted.

MikeEdgar (approved):

PR #883, opened 2023-03-13
MGDSTRM-10879 adding control plane directed pause handling

shawkins:
Allows the control plane to pause syncing to the data plane via the pause annotation https://issues.redhat.com/browse/MGDSTRM-10879

The delete flag is still set to allow for cleanup to proceed.

This usage of the annotation will not intersect with sre usage of the pause annotation, but we can of course use an entirely different annotation key if anyone objects to this reuse.

### DIFF
--- a/sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
+++ b/sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
@@ -2,12 +2,13 @@ package org.bf2.sync;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.fabric8.kubernetes.api.model.Namespace;
 import io.fabric8.kubernetes.api.model.NamespaceBuilder;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.Secret;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.kubernetes.client.dsl.base.PatchContext;
+import io.fabric8.kubernetes.client.dsl.base.PatchType;
 import io.fabric8.kubernetes.client.informers.cache.Cache;
 import io.fabric8.kubernetes.client.utils.KubernetesResourceUtil;
 import io.fabric8.kubernetes.client.utils.Serialization;
@@ -41,7 +42,6 @@ import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 
 /**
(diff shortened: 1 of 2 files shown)


sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
```diff
@@ -103,6 +103,25 @@ public void syncKafkaClusters() {
...
+                    continue;
+                }
+                if (existing != null) {
+                    if (!existing.getSpec().isDeleted()) {
+                        log.debugf("Passing just the delete flag to the ManagedKafka %s", Cache.metaNamespaceKeyFunc(remoteManagedKafka));
+                        kubeClient.resources(ManagedKafka.class)
+                                .inNamespace(existing.getMetadata().getNamespace())
+                                .withName(existing.getMetadata().getName())
+                                .patch(PatchContext.of(PatchType.JSON),
+                                        "[{\"op\": \"replace\", \"path\":\"/spec/deleted\", \"value\":true}]");
```
MikeEdgar, line 119: Why this style instead of using the POJO?

shawkins: Strategic merge is not supported.  There isn't a "good" direct way of producing the simplified json patch.  And json merge isn't great for partial documents, but may be fine here, I didn't check specifically.

machi1990 (approved):
lgtm, thanks @shawkins 

tested this PR from the control plane perspective.
let me know when the annotation key changes to something else.

MikeEdgar (approved):

sync/src/main/java/org/bf2/sync/ManagedKafkaSync.java
rareddy, line 119: little confused about the logic here, if the control plane pause is true, and if the local instance exists and is not deleted it sets "spec/delete" to true?

shawkins: You'll only do that if the remote is marked as deleted:

```
if remote paused {
  if remote not deleted {
    move to the next item to skip main reconciliation
  }
  if local exists {
    if local is not deleted {
      mark local deleted
    }
    move to the next item to skip main reconciliation
  }
} 
```

rareddy: Thanks! remote deleted and paused at the same time seems like a weird case imo.

machi1990: `deleted` is a status in the database. 
`pause-reconciliation` directly ties to whether some resources e.g TLS certificates where able to be fetched or not. These resources are fetched each time a ManagedKafka is reconciled with the Fleetshard sync irrespective of the status (wether deleted or not) of the Kafka.

rareddy: ok. Thanks for explanation.

rareddy (approved):
lgtm

biswassri (approved):

machi1990:
Thanks, will this be on the 0.35 release?

shawkins:
@machi1990 yes it will be.

machi1990:
Thanks

PR #887, opened 2023-03-23
more completely addressing sonar complaints

shawkins:
These were not completely addressed with the previous commit.


biswassri (approved):
lgtm

PR #888, opened 2023-03-23
Approve Strimzi installation immediately when CSV not present

MikeEdgar:
In addition to the case when the CRDs are not present, also approve Strimzi installation immediately when the subscription does not have a linked CSV yet. This likely only applies to test environments where the operator is installed/removed repeatedly and the CRDs are kept in the cluster.

### DIFF
--- a/operator/src/main/java/org/bf2/operator/managers/StrimziBundleManager.java
+++ b/operator/src/main/java/org/bf2/operator/managers/StrimziBundleManager.java
@@ -202,7 +202,7 @@ public class StrimziBundleManager {
 
         final String subNamespace = subscription.getMetadata().getNamespace();
         final String subName = subscription.getMetadata().getName();
-        boolean approveImmediately;
+        boolean approveImmediately = false;
 
         if (subscription.getStatus().getInstalledCSV() == null) {
             log.infof("Subscription %s/%s has no linked CSV; InstallPlan will be approved immediately", subNamespace, subName);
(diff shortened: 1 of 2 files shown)


operator/src/main/java/org/bf2/operator/managers/StrimziBundleManager.java
```diff
@@ -200,10 +200,23 @@ private Approval approveInstallation(Subscription subscription) {
...
+        boolean approveImmediately;
+
+        if (subscription.getStatus().getInstalledCSV() == null) {
+            log.infof("Subscription %s/%s has no linked CSV; InstallPlan will be approved immediately", subNamespace, subName);
+            approveImmediately = true;
+        } else if (!this.isKafkaCrdsInstalled()) {
+            log.infof("Subscription %s/%s has missing Strimzi CRDs; InstallPlan will be approved immediately", subNamespace, subName);
+            approveImmediately = true;
+        } else {
+            approveImmediately = false;
```
shawkins, line 214: Minor nit, just initialize approveImmediately as false.

shawkins (approved):
LGTM

operator/src/main/java/org/bf2/operator/managers/StrimziBundleManager.java
MikeEdgar: Done, latest commit also removes some modifiers in the unit test class Sonar flags as smells.

biswassri (approved):
LGTM

PR #889, opened 2023-03-24
[MGDSTRM-9807] Separate image-jib to active-by-default profile

MikeEdgar:
This seems to be the configuration that will allow us to skip the JIB processing when needed.

### DIFF
--- a/operator/pom.xml
+++ b/operator/pom.xml
@@ -30,6 +30,12 @@
         <dependency>
             <groupId>io.quarkiverse.operatorsdk</groupId>
             <artifactId>quarkus-operator-sdk</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-container-image-jib</artifactId>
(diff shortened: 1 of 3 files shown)


shawkins (approved):
LGTM

Since it's the same profile and we reference it from the root pom.xml, what about moving it up to root?

rareddy:
Couple questions

> I considered that but I didn't like that it would be added to all the other modules via the parent.

why we do not want for all the modules?

since we use the profile in release, the metadata file as u were saying will show up in the image anyways, then what are we gaining by putting it in profile?

MikeEdgar:
> why we do not want for all the modules?

It's only sync and operator that need it when doing a JIB image build. The others like api, common, and the test modules don't use it for anything.

> since we use the profile in release, the metadata file as u were saying will show up in the image anyways, then what are we gaining by putting it in profile?

The release profile is used for GitHub releases, but we don't use that profile for the product build.

rareddy (approved):
lgtm

PR #900, opened 2023-03-30
MGDSTRM-11040 allowing the platform to change and adding clientAccess

shawkins:
ingresscontroller configuration needs updated for the internal gcp case - the type needs to be GCP and the gcp params need to specify clientAccess: global.  In any other situation that parameter will be ignored.


operator/src/test/java/org/bf2/operator/managers/IngressControllerManagerTest.java
```diff
@@ -552,6 +555,29 @@ void testIngressControllerInternal() {
...
+                .withNewSpec()
+                .editOrNewNet().withPrivate(true).endNet()
+                .endSpec()
+                .build();
+        IngressController controller = ingressControllerManager.buildIngressController("kas", "kas", null, 1, null, null, agent);
+        assertEquals("---\n"
+                + "aws:\n"
+                + "  type: \"NLB\"\n"
+                + "gcp:\n"
+                + "  clientAccess: \"Global\"\n"
```
MikeEdgar, line 577: Presumably having entries for providers other than the one the cluster is actually using is harmless. I suppose the experience with having AWS all this time on GCP proves that.

MikeEdgar (approved):
LGTM

operator/src/test/java/org/bf2/operator/managers/IngressControllerManagerTest.java
shawkins: Yes it's an odd system.  The type has to match before the cluster ingress operator will even look at the parameters, so the aws params have been ignored this whole time on gcp.  Also the operator won't even look at the clientAccess unless the scope is internal.

rareddy (approved):
lgtm

PR #907, opened 2023-04-07
MGDSTRM-11101 delaying the creation of ingresscontrollers

shawkins:
Delays the creation of ingresscontrollers until after we have a kafka.  A delayCreate member was added to simplify keeping the existing test logic with the same expectations, rather than adding dummy kafkas.


MikeEdgar (approved):
Looks good to me.

Do you think we should also consider (in a separate task/PR) the case where the last Kafka is deleted and the ICs are unused? I was thinking we could remove them then or, at worst, when the operator is uninstalled.

shawkins:
> Do you think we should also consider (in a separate task/PR) the case where the last Kafka is deleted and the ICs are unused?

I was hoping to have clarity on the public kafka on private rosa scenario before making a change like that.

shawkins:
@grdryn @k-wall can someone from kafka integrations comment on delaying the creation of the ingress controllers https://issues.redhat.com/browse/MGDSTRM-11101

grdryn (approved):
Looks reasonable to me, thanks :+1:

PR #908, opened 2023-04-13
rough draft of horizontal-ish scaling

shawkins:


MikeEdgar:
> It does not seem like a good idea to go down the path of copying the hash functionality in Java, nor trying to time our updates to the deployment to match expectations with the OLM operator. So this is likely a dead end.

Patching the deployment(s) within the CSV might be viable. I don't think OLM will attempt to modify or restore the CSV back to the version unpacked from the bundle.

shawkins:
> Patching the deployment(s) within the CSV might be viable. 

Thanks for the suggestion to move things up a level.  

> I don't think OLM will attempt to modify or restore the CSV back to the version unpacked from the bundle.

At least locally it does not seem to.  This will put the CSV back into the installing phase temporarily - are you aware of any monitoring or other reason that would be an issue?

MikeEdgar:
> This will put the CSV back into the installing phase temporarily - are you aware of any monitoring or other reason that would be an issue?

It's possible that during the initial install the OCM status for the addon would take longer than otherwise to become Ready, but it won't impact readiness to provision a Kafka from KFM's perspective since it relies on the MKA's status for a Hybrid installation. We shouldn't have any monitoring of that at the service level.

shawkins:
@k-wall @MikeEdgar @rareddy here's an update of the draft using the csv modification approach.  I've also narrowed it to just the horizontal / replica manipulation.  The resource discussion will continue on https://issues.redhat.com/browse/MGDSTRM-9976 and https://issues.redhat.com/browse/MGDSTRM-11132 but if this pr seems acceptable, we can reintroduce this approach as an alternative as well.

PR #910, opened 2023-04-18
MGDSTRM-10557: Expose strimzi oauth metrics prometheus in order to assist investigating a slow authentication issue

k-wall:
I also needed to apply the override mechanism to the kafka container in order to set kafka container environment variables, so I can actually [turn on strimzi oauth metrics](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/README.md#configuring-the-metrics). The work for this part is a separate commit.

```
 env:
      - name: OAUTH_ENABLE_METRICS
        value: "true"
```

Here are screenshots showing the new metrics, demonstrating that the end to end is working.

Screenshot 2023-04-18 at 12 02 29

Screenshot 2023-04-18 at 12 09 23

### DIFF
--- a/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
+++ b/operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
@@ -26,6 +26,8 @@ import io.fabric8.openshift.api.model.Route;
 import io.fabric8.openshift.api.model.TLSConfigBuilder;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.quarkus.arc.DefaultBean;
+import io.strimzi.api.kafka.model.ContainerEnvVar;
+import io.strimzi.api.kafka.model.ContainerEnvVarBuilder;
 import io.strimzi.api.kafka.model.CruiseControlSpec;
 import io.strimzi.api.kafka.model.CruiseControlSpecBuilder;
 import io.strimzi.api.kafka.model.ExternalConfigurationReferenceBuilder;
@@ -98,6 +100,7 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 import static io.fabric8.kubernetes.api.model.Quantity.getAmountInBytes;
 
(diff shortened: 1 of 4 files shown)


operator/src/main/java/org/bf2/operator/operands/KafkaCluster.java
```diff
@@ -554,6 +562,15 @@ private KafkaClusterTemplate buildKafkaTemplate(ManagedKafka managedKafka, int r
         return templateBuilder.build();
     }
 
+    private List<ContainerEnvVar> buildKafkaContainerEnvVars(ManagedKafka managedKafka) {
+        var kafkaOverride = this.overrideManager.getKafkaOverride(managedKafka.getSpec().getVersions().getStrimzi());
```
k-wall, line 566: I tried approaches introducing an env var type parameter to `OperandOverride<E>`) etc but the resulting code `applyEnvironmentTo` in  less obvious that this list conversion.

operator/src/main/resources/kafka-metrics.yaml
```diff
@@ -305,3 +305,64 @@ data:
...
+          "$8": "$9"
+      - pattern: "strimzi.oauth<type=(.+), context=(.+), kind=(.+), host=\"(.+)\", path=\"(.+)\", (.+)=(.+)><>(count|totalTimeMs):"
+        name: "strimzi_oauth_$1_$8"
+        type: COUNTER
+        labels:
+          context: "$2"
+          kind: "$3"
+          host: "$4"
+          path: "$5"
+          "$6": "$7"
```
grdryn, line 338: I wonder if these three entries are redundant? The 3 patterns below look like more generalised forms that would also cover these cases?

grdryn (approved):
One small comment/question, but looks good to me.

operator/src/main/resources/kafka-metrics.yaml
k-wall: No, they won't be redundant.  They are matching different forms. The number of groups is different in each case.

MikeEdgar (approved):
LGTM :+1:

operator/src/main/resources/kafka-metrics.yaml
grdryn: I had counted the number of groups in each case, and they appeared to be the same is what I meant. :slightly_smiling_face:  It seems like the actual difference is that the above 3 entries are counters, while the below 3 are gauges. Here's a diff tool view with the above on the left, and the below on the right. Lines with differences are those in blue, with the actual differences on those lines highlighted in a slightly lighter blue:

image

